PR #1, opened 2018-03-20
New Koji model class and relationship bug fix for other models

yashvardhannanavati:


purview/models/freshmaker.py
```diff
@@ -1,11 +1,11 @@
 from neomodel import (
-    StructuredNode, UniqueIdProperty, RelationshipFrom, RelationshipTo,
```
mprahl: In the future, the relationship fixes should be in a separate commit.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    StructuredNode, StringProperty, IntegerProperty, UniqueIdProperty,
+    BooleanProperty, DateTimeProperty, FloatProperty, RelationshipTo)
+
+
+class KojiBuild(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')
+    tasks = RelationshipTo('KojiTask', 'TRIGGERS')
+    creation_event_id = IntegerProperty(required=True)
+    volume_id = IntegerProperty()
+    owner_id = IntegerProperty()
```
mprahl: This would be a good candidate for a separate node. We should have a generic user node (used for relationships with all labels) that has a relationship with the build.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    package_id = IntegerProperty(required=True)
```
mprahl: This is another good candidate to be broken out into a separate label.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    task_id = IntegerProperty(required=True)
```
mprahl: I don't know if this is necessary since we have the relationship to the tasks through the `tasks` property.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    build_id = IntegerProperty()
```
mprahl: Not sure, but the `build_id` should probably be the `UniqueIdProperty` of the label/node. What does `id_` currently stand for if it's not the `build_id`?

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    name = StringProperty()
+    owner_name = StringProperty()
```
mprahl: This can be removed once the user stuff is broken out into a separate label/node.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    package_name = StringProperty(required=True)
```
mprahl: This should go in to the package label/node.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    completion_time = DateTimeProperty()
+    start_ts = FloatProperty()
+    creation_ts = FloatProperty(required=True)
+    completion_ts = FloatProperty()
+    extra = StringProperty()
+
+
+class KojiTask(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')
+    builds = RelationshipTo('KojiBuildNode', 'TIGGERED_BY')
```
mprahl: "TIGGERED_BY" => "TRIGGERED_BY"

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    start_ts = FloatProperty()
+    state = IntegerProperty()
+    awaited = BooleanProperty(required=True)
+    label = StringProperty(required=True)
+    priority = IntegerProperty()
+    channel_id = IntegerProperty()
+    waiting = StringProperty()
+    create_time = DateTimeProperty(required=True)
+    create_ts = FloatProperty(required=True)
+    owner = IntegerProperty()
```
mprahl: This should be broken into a separate node as mentioned above.

mprahl:
@yashvardhannanavati I left some comments. I'm also wondering if it's worth keeping those foreign keys that we don't expand to individual labels/nodes. For instance `volume_id` on it's own is not very useful, we'd have to store data about the volume to make it useful. Since we likely won't have a use for information about the volumes the RPMs are in for this app, we should probably not store that.

purview/models/freshmaker.py
yashvardhannanavati: Yeap will keep that in mind!

yashvardhannanavati:
@mprahl  We can definitely break up the data into multiple nodes, but again the main concern would be that once we expose the API and a person queries for the builds, he/she will expect all the properties of a koji build. In which case, the person accessing the data will have to handle these things in their program. I think it would be better to keep it the way we receive it. Please let me know your thoughts on this

mprahl:
@yashvardhannanavati I think it's important to note that our Neo4j instance isn't trying to be a competitor to our Teeid deployment so the schema doesn't need to exactly match. Also, I don't know if we'll actually expose the API to anything but our frontend after talking with Jay. Unless there's a compelling reason to expose this, then we should avoid it so that we can break the API when we need to.

As for the foreign keys, we can keep those in there until we develop separate nodes for them when the need arises, but again, it's more data to keep track of and maintain. Whatever way we go here, we should still remove things like `owner_name` because it's unnecessary work and data duplication to maintain a relationship to a separate user node and maintain a data property on the build.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
+from neomodel import (
+    StructuredNode, StringProperty, IntegerProperty, UniqueIdProperty,
+    BooleanProperty, DateTimeProperty, FloatProperty, RelationshipTo)
+
+
+class KojiBuild(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')
+    tasks = RelationshipTo('KojiTask', 'TRIGGERS')
+    creation_event_id = IntegerProperty(required=True)
```
ralphbean: Probably... remove this?

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    volume_id = IntegerProperty()
```
ralphbean: Remove this.

purview/models/koji.py
ralphbean: Yeah, agreed.  Package as a StructuredNode of its own with a relationship.

purview/models/koji.py
ralphbean: In our call, @mprahl suggested removing this and storing the "brew build id" from the brew system in our `id_` UniqueIdProperty.

We should test that a KojiBuild with id_ 42 doesn't conflict with a FreshmakerEvent with id_ 42.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    owner_id = IntegerProperty()
+    package_id = IntegerProperty(required=True)
+    task_id = IntegerProperty(required=True)
+    build_id = IntegerProperty()
+    name = StringProperty()
```
ralphbean: Check what this actually is.  :p

Maybe remove.

ralphbean: Actually, this seems to be related to "version" and "release" below.  We probably need to keep it to construct the unique NVR of the build (this is how people talk about builds, as NVRs).

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    owner_name = StringProperty()
+    package_name = StringProperty(required=True)
+    version = StringProperty(required=True)
+    release = StringProperty(required=True)
+    epoch = StringProperty()
+    nvr = StringProperty()
+    source = StringProperty()
+    state = IntegerProperty(required=True)
+    volume_name = StringProperty()
```
ralphbean: This can .. probably go.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    start_time = DateTimeProperty()
+    creation_time = DateTimeProperty(required=True)
+    completion_time = DateTimeProperty()
+    start_ts = FloatProperty()
```
ralphbean: Remove either the "ts" or the "time" version of this so we only have one.  Pick whichever is most natural.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+
+class KojiTask(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')
+    builds = RelationshipTo('KojiBuildNode', 'TIGGERED_BY')
+    children = RelationshipTo('KojiTask', 'CHILD')
+    parents = RelationshipTo('KojiTask', 'PARENT')
+    weight = FloatProperty(required=True)
+    completion_time = DateTimeProperty()
+    start_time = DateTimeProperty()
+    start_ts = FloatProperty()
```
ralphbean: Same here about timestamps vs time objects.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    source = StringProperty()
+    state = IntegerProperty(required=True)
+    volume_name = StringProperty()
+    start_time = DateTimeProperty()
+    creation_time = DateTimeProperty(required=True)
+    completion_time = DateTimeProperty()
+    start_ts = FloatProperty()
+    creation_ts = FloatProperty(required=True)
+    completion_ts = FloatProperty()
+    extra = StringProperty()
```
mprahl: Let's parse this in the scraper and a `scratch` property on the KojiBuild node.

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    id_ = UniqueIdProperty(db_property='id')
+    builds = RelationshipTo('KojiBuildNode', 'TIGGERED_BY')
+    children = RelationshipTo('KojiTask', 'CHILD')
+    parents = RelationshipTo('KojiTask', 'PARENT')
+    weight = FloatProperty(required=True)
+    completion_time = DateTimeProperty()
+    start_time = DateTimeProperty()
+    start_ts = FloatProperty()
+    state = IntegerProperty()
+    awaited = BooleanProperty(required=True)
```
ralphbean: Can probably remove this?

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    label = StringProperty(required=True)
+    priority = IntegerProperty()
+    channel_id = IntegerProperty()
+    waiting = StringProperty()
```
ralphbean: This can.. probably go?

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    builds = RelationshipTo('KojiBuildNode', 'TIGGERED_BY')
+    children = RelationshipTo('KojiTask', 'CHILD')
+    parents = RelationshipTo('KojiTask', 'PARENT')
+    weight = FloatProperty(required=True)
+    completion_time = DateTimeProperty()
+    start_time = DateTimeProperty()
+    start_ts = FloatProperty()
+    state = IntegerProperty()
+    awaited = BooleanProperty(required=True)
+    label = StringProperty(required=True)
```
mprahl: Remove this

purview/models/koji.py
```diff
@@ -0,0 +1,55 @@
...
+    priority = IntegerProperty()
+    channel_id = IntegerProperty()
+    waiting = StringProperty()
+    create_time = DateTimeProperty(required=True)
+    create_ts = FloatProperty(required=True)
+    owner = IntegerProperty()
+    host_id = IntegerProperty()
```
ralphbean: Let's ditch this one too.

purview/models/koji.py
mikebonnet: `name` and `package_name` are the same thing. If Package is being moved into its own node, then these can both be removed. Unless there's value to duplicating the name in the KojiBuild (for query/filter reasons)?

purview/models/koji.py
mikebonnet: `scratch` should be a property of `KojiTask`. A `KojiBuild` will never be created for a scratch task. This `extra` property is actually extra metadata in json format, usually provided by a `Content Generator` when it imports a build from another system. I would suggest retaining it, it might be useful for handling container and Middleware builds.

purview/models/koji.py
```diff
@@ -0,0 +1,36 @@
...
+    id_ = UniqueIdProperty(db_property='id')
+    tasks = RelationshipTo('KojiTask', 'TRIGGERS')
+    owner = RelationshipTo('.user.User', "OWNED_BY")
+    epoch = StringProperty()
+    state = IntegerProperty(required=True)
+    creation_time = DateTimeProperty(required=True)
+    start_time = DateTimeProperty()
+    completion_time = DateTimeProperty()
+    extra = StringProperty()
+    name = StringProperty()
```
mprahl: Can you make this required?

purview/models/koji.py
```diff
@@ -0,0 +1,36 @@
...
+    version = StringProperty(required=True)
+    release = StringProperty(required=True)
+    nvr = StringProperty()
```
mprahl: Can you make this required?

purview/models/koji.py
```diff
@@ -0,0 +1,36 @@
...
+    id_ = UniqueIdProperty(db_property='id')
+    builds = RelationshipTo('KojiBuild', 'TRIGGERED_BY')
+    children = RelationshipTo('KojiTask', 'CHILD')
+    parents = RelationshipTo('KojiTask', 'PARENT')
+    owner = RelationshipTo('.user.User', "OWNED_BY")
+    weight = FloatProperty()
+    create_time = DateTimeProperty(required=True)
+    start_time = DateTimeProperty()
+    completion_time = DateTimeProperty()
+    state = IntegerProperty()
```
mprahl: I think we can make this required as well.

mprahl:
Nice :shipit:

PR #3, opened 2018-03-21
Add the dist-git models

mprahl:


purview/models/distgit.py
```diff
@@ -0,0 +1,26 @@
...
+
+from neomodel import (
+    StructuredNode, UniqueIdProperty, RelationshipTo, StringProperty,
+    DateTimeProperty)
+
+
+class DistgitCommit(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')  # commit_id
+    owner = RelationshipTo('.user.User', 'OWNED_BY')  # author
+    committer = RelationshipTo('.user.User', 'OWNED_BY')
```
mprahl: @ralphbean do you think the `committer` is unnecessary data? I think we only care about the author of the commit.

purview/models/distgit.py
```diff
@@ -0,0 +1,26 @@
...
+    pushes = RelationshipTo('DistgitPush', 'PUSHED_BY')
```
mprahl: @yashvardhannanavati  and @ralphbean what are your thoughts on the wording here for the relationship?

purview/models/distgit.py
yashvardhannanavati: Do you think adding these relationships to the User class as well and including that in this PR is a good idea?

purview/models/distgit.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from neomodel import (
+    StructuredNode, UniqueIdProperty, RelationshipTo, StringProperty,
+    DateTimeProperty)
+
+
+class DistgitCommit(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')  # commit_id
+    owner = RelationshipTo('.user.User', 'OWNED_BY')  # author
```
yashvardhannanavati: If this is the author of the commit, do you think "AUTHORED_BY" makes sense for this relationship?

mprahl: I do but I tried to reuse `OWNED_BY` for simplicity in our queries.

ralphbean: Yeah, I guess this is the "normalized" author relationship.  :)

purview/models/distgit.py
ralphbean: If you keep the committer, you need to also keep the author *as* the author in addition to the author *as* the owner.

Let's keep all three.

purview/models/distgit.py
ralphbean: Hm, use `PUSHED_IN` since "by" makes me thing of a person who pushed.

purview/models/distgit.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from neomodel import (
+    StructuredNode, UniqueIdProperty, RelationshipTo, StringProperty,
+    DateTimeProperty)
+
+
+class DistgitCommit(StructuredNode):
```
ralphbean: Can you construct a URL back to the commit with the data you have here?  I think you're missing repo name, namespace, etc.

mprahl:
@yashvardhannanavati and @ralphbean thanks for the review. I made the suggested changes. I made all relationship property names plural because neomodel returns them as a list even though all those relationships will only be tied to one node. What do you guys think of this approach?

yashvardhannanavati:
@mprahl Sounds good to me!

purview/models/distgit.py
mprahl: The repo name is called `module` in the database. Maybe it's worth renaming it here for clarity. I'll double check on the namespace.

mprahl: @ralphbean so looking at the gitbz code, the namespace is not stored... let's talk tomorrow about workarounds.

mprahl:
I added Bugzilla references. Can you please review @yashvardhannanavati and @ralphbean?

purview/models/bugzilla.py
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from neomodel import StructuredNode, UniqueIdProperty, RelationshipTo
+
+
+class BugzillaBug(StructuredNode):
```
mprahl: This will be filled out more later but this is the least we need for dist-git.

purview/models/distgit.py
lubomir: When namespaces were added, we did not update gitbz (an oversight really, but it was not clear if it's actually used anywhere). It would be really to modify it to store the module as `namespace/name`. Adding another column for the namespace would also be an option.

mikebonnet: I prefer the idea of a separate namespace column. If we're touching the data model, it would be a good opportunity to add `author_email` and `committer_email` as well.

lubomir: I opened COMPOSE-2409 to track adding that.

mprahl:
I'll close this PR and reopen once I have the scraper ready.

PR #5, opened 2018-03-22
Add Teiid defaults to the BaseScraper class

mprahl:


scrapers/base.py
```diff
@@ -5,6 +5,9 @@ class BaseScraper(object):
     """
     Base scraper class to standardize the main scraper functionality
     """
+    teiid_hostname = 'virtualdb.engineering.redhat.com'
```
yashvardhannanavati: Can we just have this as teiid_host instead of teiid_hostname ? Just for uniformity with the teiid class

mprahl: Sure

mprahl:
Fixed

PR #9, opened 2018-03-23
Add a timestamp parsing function

mprahl:
This will be useful for when we parse the `since` input for our scrapers.

### DIFF
--- a/purview/utils/general.py
+++ b/purview/utils/general.py
@@ -1,0 +1,36 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+import re
+import logging
+from datetime import datetime
+
+log = logging.getLogger(__name__)
+
+
+def timestamp_to_datetime(timestamp):
+    """
+    Parses a timestamp and return a datetime object
+    :param timestamp: a string of a timestamp
+    :return: a datetime object
+    """
+    log.debug('Trying to parse the timestamp "{0}"'.format(timestamp))
+    error_msg = 'The timestamp "{0}" is an invalid format'.format(timestamp)
+    combinations = (
+        (r'^(?P<datetime>\d{4}-\d{1,2}-\d{1,2} \d{1,2}:\d{1,2}:\d{1,2})(?:\.\d+)?$',
+         '%Y-%m-%d %H:%M:%S'),
+        (r'^(?P<datetime>\d{4}-\d{1,2}-\d{1,2})$', '%Y-%m-%d'),
+        # ISO 8601 format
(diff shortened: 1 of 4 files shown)


purview/utils/general.py
```diff
@@ -0,0 +1,29 @@
...
+def timestamp_to_datetime(timestamp):
+    """
+    Parses a timestamp and return a datetime object
+    :param timestamp: a string of a timestamp
+    :return: a datetime object
+    """
+    log.debug('Trying to parse the timestamp "{0}"'.format(timestamp))
+    combinations = (
+        (r'^(?P<datetime>\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2})(?:\.\d+)?$', '%Y-%m-%d %H:%M:%S'),
+        (r'^(?P<datetime>\d{4}-\d{2}-\d{2})$', '%Y-%m-%d'),
```
yashvardhannanavati: Just one concern, I don't think it will match something like 2016-3-3

mprahl: Good point. I'll tweak the regex.

purview/utils/general.py
```diff
@@ -0,0 +1,35 @@
...
+    :param timestamp: a string of a timestamp
+    :return: a datetime object
+    """
+    log.debug('Trying to parse the timestamp "{0}"'.format(timestamp))
+    error_msg = 'The timestamp "{0}" is an invalid format'.format(timestamp)
+    combinations = (
+        (r'^(?P<datetime>\d{4}-\d{1,2}-\d{1,2} \d{1,2}:\d{1,2}:\d{1,2})(?:\.\d+)?$',
+         '%Y-%m-%d %H:%M:%S'),
+        (r'^(?P<datetime>\d{4}-\d{1,2}-\d{1,2})$', '%Y-%m-%d'),
+        (r'^(?P<datetime>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})(?:\.\d+)?(?:Z|[-+]00(?::00)?)?$',
```
yashvardhannanavati: Can we do the same here as well? Or am I missing something?

mprahl: Not there because I'm following the ISO-8601 spec there.

yashvardhannanavati: ah, sounds good!

tox.ini
```diff
@@ -1,11 +1,15 @@
 [tox]
-envlist = flake8
+envlist = flake8,py27,p36
```
yashvardhannanavati: (absolutely optional) Space missing after comma? :P

mprahl: Sure, let me try it!

mprahl: Fixed

yashvardhannanavati:
+1

PR #11, opened 2018-03-26
Added KojiTag model and it's relationships

yashvardhannanavati:


yashvardhannanavati:
Assuming we will only store active tags

purview/models/koji.py
```diff
@@ -34,3 +35,10 @@ class KojiTask(StructuredNode):
     priority = IntegerProperty()
     arch = StringProperty(required=True)
     method = StringProperty(required=True)
+
+
+class KojiTag(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')
+    creator = RelationshipTo('.user.User', 'CREATED_BY')
```
mprahl: Can you use `owners` here for consistency with our other models? Then the relationship name in Neo4j should be `OWNED_BY`.

mprahl: On second thought, maybe we don't need this data at all.

yashvardhannanavati: Sure, I can remove that relationship

mprahl: Okay, if you do can you remove the inverse as well?

yashvardhannanavati: Absolutely!, now please enjoy your sunday night. you can review this tomorrow :)

purview/models/koji.py
```diff
@@ -34,3 +35,9 @@ class KojiTask(StructuredNode):
     priority = IntegerProperty()
     arch = StringProperty(required=True)
     method = StringProperty(required=True)
+
+
+class KojiTag(StructuredNode):
+    id_ = UniqueIdProperty(db_property='id')
+    builds = RelationshipTo('KojiBuild', 'TAGGED_TO')
```
mprahl: Based on our discussion, this should likely be called `CONTAINED_BY` for more generic relationship names.

purview/models/koji.py
```diff
@@ -9,6 +9,7 @@ class KojiBuild(StructuredNode):
     id_ = UniqueIdProperty(db_property='id')
     tasks = RelationshipTo('KojiTask', 'TRIGGERS')
     owner = RelationshipTo('.user.User', 'OWNED_BY')
+    tags = RelationshipTo('KojiTag', 'TAGGED_BY')
```
mprahl: Based on our discussion, this should likely be called `CONTAINED_BY` for more generic relationship names.

mprahl:
:+1:

PR #17, opened 2018-03-26
Added Bugzilla relationships to User model

yashvardhannanavati:


purview/models/user.py
```diff
@@ -9,5 +9,9 @@ class User(StructuredNode):
     username = UniqueIdProperty()
     koji_builds = RelationshipTo('.koji.KojiBuild', 'OWNS')
     koji_tasks = RelationshipTo('.koji.KojiTask', 'OWNS')
+    bugzilla_bugs = RelationshipTo('.bugzilla.BugzillaBug', 'OWNS')
+    bugzilla_assignees = RelationshipTo('.bugzilla.BugzillaBug', 'ASSIGNED')
```
mprahl: I think the grammar should be something like `bugs_assigned_to` instead

purview/models/user.py
```diff
@@ -9,5 +9,9 @@ class User(StructuredNode):
...
+    bugzilla_reporters = RelationshipTo('.bugzilla.BugzillaBug', 'REPORTS')
```
mprahl: Same as above

purview/models/user.py
```diff
@@ -9,5 +9,9 @@ class User(StructuredNode):
...
+    bugzilla_qa_contacts = RelationshipTo('.bugzilla.BugzillaBug', 'QA')
```
mprahl: Same as above

purview/models/user.py
mprahl: `REPORTS` should also be past-tense since it happened in the past.

mprahl:
:+1:

PR #18, opened 2018-03-26
Added bugzilla scraper

yashvardhannanavati:


scrapers/bugzilla.py
```diff
@@ -0,0 +1,113 @@
...
+# To support user names in other languages
+try:
+    # Py3 requires this import
+    from importlib import reload
+except ImportError:
+    # Py2 has reload builtin
+    log.debug("Cannot import importlib")
+
+reload(sys)
+sys.setdefaultencoding('utf-8')
```
mprahl: This is very hacky. Can you not just decode the string individually?

scrapers/bugzilla.py
```diff
@@ -0,0 +1,98 @@
...
+from purview.models.bugzilla import BugzillaBug
+from purview.models.user import User
+import purview.utils.general as utils
+from purview import log
+
+
+class BugzillaScraper(BaseScraper):
+
+    def run(self, since=None):
+        log.info('Starting initial load for Bugzilla bugs')
```
mprahl: "for" => "of"

scrapers/bugzilla.py
```diff
@@ -0,0 +1,98 @@
...
+
+        bugs = self.get_bugzilla_bugs(since)
+        log.info('Successfully fetched {0} bugs from teiid'.format(len(bugs)))
+        self.update_neo4j(bugs)
+        log.info('Initial load of Bugzilla bugs complete!')
+
+    def get_bugzilla_bugs(self, start_date):
+        # SQL query to fetch all (Red Hat) bugs from start date until now
+        log.info('Getting all Bugzilla bugs since {0}'.format(start_date))
+        sql_query = """
```
mprahl: As discussed, you should be able to left join all the users here so you only need to do one SQL query.

yashvardhannanavati: Done!

scrapers/bugzilla.py
yashvardhannanavati: Done!

scrapers/bugzilla.py
```diff
@@ -0,0 +1,105 @@
...
+            WHERE classifications.name = 'Red Hat' AND bugs.creation_ts > '{}'
+            """.format(start_date)
+
+        return self.teiid.query(sql=sql_query, retry=3)
+
+    def create_user_node(self, email, name):
+        # if email is a Red Hat email address, username is
+        # same as domain name prefix in the email address
+        # else store email as username
+        print(name)
```
mprahl: Can you remove this please?

scrapers/bugzilla.py
```diff
@@ -0,0 +1,105 @@
...
+        if email.split("@")[1] == 'redhat.com':
```
mprahl: Can you use single quotes to match the rest of the code please?

scrapers/bugzilla.py
```diff
@@ -0,0 +1,105 @@
...
+            username = email.split("@")[0]
```
mprahl: Same as above

scrapers/bugzilla.py
```diff
@@ -0,0 +1,108 @@
+# SPDX-License-Identifier: GPL-3.0+
+from __future__ import unicode_literals
+
+from builtins import bytes
+
+from scrapers.base import BaseScraper
+from purview.models.bugzilla import BugzillaBug
+from purview.models.user import User
+import purview.utils.general as utils
```
mprahl: This import will be confusing after #20 is merged because I'm adding a `scraper.utils` file.

scrapers/bugzilla.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class BugzillaScraper(BaseScraper):
+
+    def run(self, since=None):
+        log.info('Starting initial load of Bugzilla bugs')
+        # Initialize a start date from which all builds must be fetched
+        # If no input is given by the user, fetch builds from the past two years
+        if since is None:
+            since = self.default_since
```
mprahl: Can you not overwrite the `since` parameter for clarity? How about `since_dt`?

scrapers/bugzilla.py
```diff
@@ -0,0 +1,108 @@
...
+        # same as domain name prefix in the email address
+        # else store email as username
+        if email.split('@')[1] == 'redhat.com':
+            username = email.split('@')[0]
+        else:
+            username = email
+
+        user = User.create_or_update(dict(
+            username=username,
+            name=bytes(name, 'utf-8').decode(),
```
mprahl: Let's remove the name here since this isn't the canonical source for names (it's a free-form setting field).

mprahl:
:+1:

PR #19, opened 2018-03-26
Use a global logger so the logging level can be set at the package level

mprahl:
Running the scraper script in master was very noisy (it included debug logs from requests, nemodel, etc.). This should fix that.

### DIFF
--- a/scripts/scrape.py
+++ b/scripts/scrape.py
@@ -10,8 +10,9 @@ sys.path.insert(1, os.path.abspath(os.path.join(sys.path[0], '..')))
 
 import scrapers  # noqa: E402
 
-log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


scripts/scrape.py
```diff
@@ -10,8 +10,9 @@
 
 import scrapers  # noqa: E402
 
-log = logging.getLogger(__name__)
-logging.basicConfig(level=logging.DEBUG)
+logging.basicConfig()
+log = logging.getLogger('purview')
+log.setLevel(logging.DEBUG)
```
yashvardhannanavati: you can actually include logging level in the logging.basicConfig() statement

mprahl: If I do that though, then all other loggers will inherit that log level.

yashvardhannanavati:
+1

PR #20, opened 2018-03-26
Add the dist-git models and scraper

mprahl:
This relies on #19 which is the first commit in this PR, so review #19 first.

Please note that it'd be easier to review this commit by commit.


purview/models/distgit.py
```diff
@@ -0,0 +1,49 @@
...
+    log_message = StringProperty()
+    sha = StringProperty(required=True)
+    authors = RelationshipTo('.user.User', 'AUTHORED_BY')
+    branches = RelationshipTo('DistGitBranch', 'CONTAINED_BY')
+    committers = RelationshipTo('.user.User', 'OWNED_BY')
+    owners = RelationshipTo('.user.User', 'OWNED_BY')
+    pushes = RelationshipTo('DistGitPush', 'PUSHED_IN')
+    related_bugs = RelationshipTo('.bugzilla.BugzillaBug', 'RELATES_TO')
+    repos = RelationshipTo('DistGitRepo', 'CONTAINED_BY')
+    resolved_bugs = RelationshipTo('.bugzilla.BugzillaBug', 'RESOLVES')
```
mprahl: Should this be `RESOLVED` instead of `RESOLVES`? Same question for the one below.

purview/models/bugzilla.py
```diff
@@ -11,16 +11,19 @@ class BugzillaBug(StructuredNode):
     assignees = RelationshipTo('.user.User', 'ASSIGNED_TO')
     reporters = RelationshipTo('.user.User', 'REPORTED_BY')
     qa_contacts = RelationshipTo('.user.User', 'QA_BY')
-    severity = StringProperty(required=True)
-    status = StringProperty(required=True)
-    creation_time = DateTimeProperty(required=True)
-    modified_time = DateTimeProperty(required=True)  # delta_ts
-    priority = StringProperty(required=True)
+    related_by_commits = RelationshipTo('.distgit.DistGitCommit', 'RELATES_TO')
```
yashvardhannanavati: I was wondering if the relationship name for this should be in past tense too..something like "RELATED_TO"

mprahl: I'm torn on this myself since everything else is past-tense but technically, we are recording things that are related to each other, so I think `RELATES_TO` is correct since they are currently relating to each other. @ralphbean is the expert on grammar. Care to weigh in here? :stuck_out_tongue:

ralphbean: The difference here is not actually past/present tense - but is active/passive voice.

- The bug relates to the commit.
- The bug is related to the commit.

These are both present tense, but the first is active voice and the second is passive (in the second the bug isn't "doing" the relating, it is more like a passive victim of the relationship).

Differences of voice only really make a difference in prose.  Either is fine here.  If you need a decision, I guess use "RELATED_TO" only because the "*ED_*" string will be easier to remember with all the other "*ED_*" strings floating around.

mprahl: Okay, going with "ED" for consistency with the other relationships.

scrapers/distgit.py
```diff
@@ -0,0 +1,171 @@
...
+
+class DistGitScraper(BaseScraper):
+    cgit_url = 'http://pkgs.devel.redhat.com/cgit/'
+
+    def run(self, since=None):
+        if since is None:
+            dt_since = self.default_since
+        else:
+            dt_since = timestamp_to_datetime(since)
+        sql = """\
```
yashvardhannanavati: This could go in a different function

scrapers/distgit.py
```diff
@@ -0,0 +1,171 @@
...
+                else:
+                    # If the email isn't a Red Hat email address, then use the whole email address
+                    # as the username. This should only happen with erroneous git configurations.
+                    people['{0}_username'.format(person)] = people['{0}_email'.format(person)]
+            # Clear soup from RAM since we no longer need it
+            del soup
+
+            log.debug('Creating nodes associated with commit {0} and push {1}'.format(
+                result['commit_id'], result['push_id']))
+            repo = DistGitRepo.get_or_create({
```
yashvardhannanavati: Also, you could have a separate method which just uploads data to neo4j

scrapers/distgit.py
```diff
@@ -0,0 +1,171 @@
...
+            branch.commits.connect(commit)
+            branch.pushes.connect(push)
+
+            push.owners.connect(pusher)
+            push.pushers.connect(pusher)
+            push.commits.connect(commit)
+            push.branches.connect(branch)
+            push.repos.connect(repo)
+
+            commit.owners.connect(author)
```
yashvardhannanavati: Just one suggestion, if there are any attributes that can be null at times, you might want to adjust the position in which you define your relationships. For eg, in Koji, if there is no task_id for the build, then I wanted to skip the remaining loop, so I defined the relationships between the 'user' and 'builds' before I checked for a task_id. 

I don't know if something like this applies to this scraper :)

yashvardhannanavati:
Otherwise LGTM!

scrapers/distgit.py
mprahl: Thanks for the reminder. I don't think there are any null values I'll come across but I'm running the scraper now and will let go all the way if time permits to ensure that is the case.

mprahl:
@yashvardhannanavati okay, this is ready for review now.

scrapers/distgit.py
```diff
@@ -0,0 +1,199 @@
...
+import re
+
+from builtins import bytes
+from bs4 import BeautifulSoup
+from requests import ConnectionError
+
+from scrapers.base import BaseScraper
+from scrapers.utils import retry_session
+from purview.utils.general import timestamp_to_datetime
+from purview.models.distgit import DistGitCommit, DistGitPush, DistGitBranch, DistGitRepo
```
yashvardhannanavati: (optional) : Order as discussed for the model file

scrapers/distgit.py
```diff
@@ -0,0 +1,199 @@
...
+                continue
+
+            log.debug('Creating nodes associated with commit ID {0} and push ID {1}'.format(
+                result['commit_id'], result['push_id']))
+            repo = DistGitRepo.get_or_create({
+                'namespace': repo_info['namespace'],
+                'name': result['module']
+            })[0]
+            branch_name = result['ref'].rsplit('/', 1)[1]
+            branch = DistGitBranch.get_or_create(
```
yashvardhannanavati: Maybe we should keep all the stuff related to uploading to Neo4j in one method just like in koji and bugzilla scrapers? Let's keep run() small

PR #23, opened 2018-03-28
Add Freshmaker scraper

yashvardhannanavati:


requirements.txt
```diff
@@ -1 +1,2 @@
 neomodel
+urllib3
```
mprahl: python-requests should already include urllib3 and that is in the `scraper-requirements.txt` file.

scrapers/freshmaker.py
```diff
@@ -0,0 +1,84 @@
...
+from __future__ import unicode_literals
+import urllib3
+
+from scrapers.base import BaseScraper
+from purview.models.freshmaker import FreshmakerEvent, ContainerBuilds
+from scrapers.utils import retry_session
+from purview.utils.general import timestamp_to_datetime
+from purview import log
+
+urllib3.disable_warnings()
```
mprahl: Let's remove this because we don't want to have `verify=False` in our production code. I'll help you get around this.

scrapers/freshmaker.py
```diff
@@ -0,0 +1,84 @@
...
+class FreshmakerScraper(BaseScraper):
+
+    def run(self, since=None):
+        """
+        Main function that runs the Freshmaker scraper
+        :param since: a string representing a datetime to start scraping data from
+        :return: None
+        """
+        if since:
+            log.info('Start date: {0} does not apply to Freshmaker scraper!'.format(since))
```
mprahl: Maybe `log.warn` here?

mprahl: How about a message like `Ignoring the since parameter since it doesn't apply to the Freshmaker scraper.`

scrapers/base.py
```diff
@@ -26,6 +26,8 @@ def __init__(self, teiid_user=None, teiid_password=None, kerberos=False, neo4j_u
         :kwarg neo4j_server: a string of the FQDN of the Neo4j server
         :return: None
         """
+        # Defining URL for Freshmaker API
+        self.freshmaker_url = "https://freshmaker.engineering.redhat.com/api/1/events/?per_page=100"
```
mprahl: This should be on the Freshmaker scraper since other scrapers shouldn't need it. See what I did in the dist-git scraper as an example.

scrapers/freshmaker.py
```diff
@@ -0,0 +1,84 @@
...
+        log.info('Starting initial load of Freshmaker events')
+        # Initialize session and base url
+        session = retry_session()
+        fm_url = self.freshmaker_url
+        self.query_api_and_update_neo4j(session, fm_url)
+        log.info('Initial load of Freshmaker events complete!')
+
+    def query_api_and_update_neo4j(self, session, fm_url):
+        """
```
mprahl: This doc string is inaccurate

scrapers/freshmaker.py
```diff
@@ -0,0 +1,84 @@
...
+        Scrapes Freshmaker API and uploads data to Neo4j
+        :param start_date: a string representing a datetime to start scraping data from
+        :return: None
+        """
+
+        while True:
+            log.debug('Querying {0}'.format(fm_url))
+            rv_json = session.get(fm_url, verify=False).json()
```
mprahl: Remove `verify=False` since this is a bad idea in production code. I'll show you how to work around this.

scrapers/freshmaker.py
```diff
@@ -0,0 +1,84 @@
...
+        if since:
+            log.info('Start date: {0} does not apply to Freshmaker scraper!'.format(since))
+        log.info('Starting initial load of Freshmaker events')
+        # Initialize session and base url
+        session = retry_session()
+        fm_url = self.freshmaker_url
+        self.query_api_and_update_neo4j(session, fm_url)
+        log.info('Initial load of Freshmaker events complete!')
+
+    def query_api_and_update_neo4j(self, session, fm_url):
```
mprahl: The session could actually be defined in this function so that you don't have to pass it in as a parameter.

mprahl: `fm_url` shouldn't be passed in since it's a class variable. You can just access it using `self.freshmaker_url`.

scrapers/freshmaker.py
```diff
@@ -0,0 +1,84 @@
...
+                        type_name=build['type_name'],
+                        url=build['url']
+                    )
+                    if build['time_completed']:
+                        cb_params['time_completed'] = timestamp_to_datetime(
+                            build['time_completed'])
+                    cb = ContainerBuilds.create_or_update(cb_params)[0]
+                    cb.triggered_by_freshmaker_event.connect(event)
+                    event.triggers_container_builds.connect(cb)
+
```
mprahl: We'll need to create the relationships with the Errata advisories too. We can do this in this PR after we merge #25 or we can do it in a separate PR.

PR #27, opened 2018-04-02
Add the ability to run all the scrapers in one command

mprahl:


scrapers/__init__.py
```diff
@@ -1,7 +1,10 @@
...
-from scrapers.errata import ErrataScraper  # noqa: F401
-from scrapers.freshmaker import FreshmakerScraper  # noqa: F401
+from scrapers.koji import KojiScraper
+from scrapers.distgit import DistGitScraper
+from scrapers.bugzilla import BugzillaScraper
+from scrapers.errata import ErrataScraper
+from scrapers.freshmaker import FreshmakerScraper
+
+# Define the scrapers here in the order that they should be run
+all_scrapers = (BugzillaScraper, DistGitScraper, KojiScraper, ErrataScraper, FreshmakerScraper)
```
ralphbean: Does order matter?

mprahl: We've written it in a way that it doesn't matter what order you run it in, but this order seems to make logical sense to me.

PR #32, opened 2018-04-04
Workaround for an issue encounted when using BeautifulSoup4 on EL7

mprahl:
The version of BeautifulSoup4 in EPEL7 doesn't return anything when using the `string` keyword on `soup.find`.

### DIFF
--- a/scrapers/distgit.py
+++ b/scrapers/distgit.py
@@ -196,7 +196,17 @@ class DistGitScraper(BaseScraper):
         soup = BeautifulSoup(cgit_result.text, 'html.parser')
         rv = {'namespace': namespace}
         for person in ('author', 'committer'):
-            td_text = soup.find('th', string=person).next_sibling.text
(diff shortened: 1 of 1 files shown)


scrapers/distgit.py
```diff
@@ -196,7 +196,17 @@ def get_repo_info(self, repo, commit):
         soup = BeautifulSoup(cgit_result.text, 'html.parser')
         rv = {'namespace': namespace}
         for person in ('author', 'committer'):
-            td_text = soup.find('th', string=person).next_sibling.text
+            # Workaround for BS4 in EL7 since `soup.find('th', string=person)` doesn't work in
+            # that environment
+            th_tags = soup.find_all('th')
+            td_text = None
```
yashvardhannanavati: I am wondering if this assignment here is necessary

yashvardhannanavati:
Other than that LGTM!

scrapers/distgit.py
yashvardhannanavati: ah, never mind. I get it

PR #33, opened 2018-04-04
Lay boilerplate for the API

yashvardhannanavati:


purview/api/v1/__init__.py
```diff
@@ -0,0 +1,27 @@
+# SPDX-License-Identifier: GPL-3.0+
+from flask import Blueprint
+
+from purview.api.v1.error import (bad_request, not_found, ValidationError)
+
+api_v1 = Blueprint('api_v1', __name__)
```
mprahl: We won't gain anything by having blueprints for our app since there aren't sections of our Flask app (we only have an API). This adds complexity that we don't need.

purview/api/v1/logger.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def init_logging(app):
+    log_level = logging.DEBUG if app.debug else logging.INFO
+    log_to_stdout(level=log_level)
+    # In general we want to see everything from our own code,
+    # but not detailed debug messages from third-party libraries.
+    # Note that the log level on the handler above controls what
+    # will actually appear on stdout.
+    logging.getLogger().setLevel(logging.INFO)
```
mprahl: We likely don't want to see logs from third-party libraries but we can keep this here for now.

purview/api/v1/error.py
```diff
@@ -0,0 +1,34 @@
...
+def bad_request(message):
+    response = jsonify(
+        {'status': 400,
+         'error': 'bad request',
+         'message': message})
+    response.status_code = 400
+    return response
+
+
+def forbidden(message):
```
mprahl: We likely don't need this since our API won't require auth at first.

mprahl:
Can you also add these changes to `requirements.txt`? Once you write that about API endpoint, can you write a unit test for it?

purview/api/v1.py
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: GPL-3.0+
+
```
mprahl: Can you add `from __future__ import unicode_literals`?

purview/api/v1.py
```diff
@@ -0,0 +1,19 @@
...
+import pkg_resources
+
+from purview import app
+
+
+@app.route('/about')
+def version_info():
```
mprahl: Can you call this function `about`?

purview/api/v1.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Extract version info from setup.py and display it on /about endpoint
+    :return: version info
+    """
+    try:
+        VERSION = pkg_resources.get_distribution('purview').version
+    except pkg_resources.DistributionNotFound:
+        VERSION = 'unknown (Please use setup.py to install purview)'
+
+    return 'Version: {0}'.format(VERSION)
```
mprahl: Can you make this JSON like `{'version': version}`

purview/api/v1.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@app.route('/about')
+def version_info():
+    """
+    Extract version info from setup.py and display it on /about endpoint
+    :return: version info
+    """
+    try:
+        VERSION = pkg_resources.get_distribution('purview').version
```
mprahl: This should be lowercase

purview/api/v1.py
```diff
@@ -0,0 +1,19 @@
...
+    except pkg_resources.DistributionNotFound:
+        VERSION = 'unknown (Please use setup.py to install purview)'
```
mprahl: Probably shouldn't expose this to the user in the event this occurs in prod some how :laughing:. How about just `unknown`?

purview/app.py
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from flask import Flask
+
+from purview.logger import init_logging
+
+
+def create_app():
+    app = Flask(__name__)
+    # Initialize logging
```
mprahl: Bike-shed: This comment isn't useful since it's clear this is happening in the line below.

purview/base.py
```diff
@@ -0,0 +1,17 @@
...
+
+import json
+
+
+class BaseAPI(object):
+    """
+    Base API class to standardize the main API functionality
+    """
+
+    def serialize(self, node_obj):
```
mprahl: This should be a function on the base class our StructuredNode classes inherit from.

purview/logger.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+import logging
+import sys
+
+
+# Inspired from waiverdb
+def log_to_stdout(level=logging.INFO):
+    fmt = '%(asctime)s [pid %(process)5d] %(name)s %(levelname)s %(message)s'
```
mprahl: I don't think we care about the pid (`[pid %(process)5d]`) in the log statements. It's just extra noise.

purview/logger.py
```diff
@@ -0,0 +1,25 @@
...
+    datefmt = '%Y-%m-%d %H:%M:%S'
+    stream_handler = logging.StreamHandler(sys.stdout)
+    stream_handler.setLevel(level)
+    stream_handler.setFormatter(logging.Formatter(fmt=fmt, datefmt=datefmt))
+    logging.getLogger().addHandler(stream_handler)
```
mprahl: Shouldn't we specifically target the purview logger here?

purview/models/__init__.py
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: GPL-3.0+
+# Import all models here so that the base api class can find them
+from models.koji import KojiBuild, KojiTask, KojiTag
```
mprahl: I don't think we need all these imports based on my previous comment

tests/test_api_v1.py
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+import unittest
+
+from purview.api.v1 import app
+
+
+class TestAPI(unittest.TestCase):
```
mprahl: Can we just use pytest style tests without any classes?

tests/test_api_v1.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def setUp(self):
+        self.app = app.test_client()
+
+    def tearDown(self):
+        pass
+
+    def test_about_endpoint(self):
+        rv = self.app.get('/about')
+        assert b'Version:' in rv.data
```
mprahl: Instead of using bytes, you should probably decode `rv.data` instead.

mprahl:
You are missing `pytest` in the test requirements file.

tests/test_api_v1.py
yashvardhannanavati: I saw pytest docs using bytes, hence I used it.

yashvardhannanavati: Sure. I will change it

tests/test_api_v1.py
```diff
@@ -0,0 +1,18 @@
...
+
+import unittest
+
+from purview.api.v1 import app
+
+
+class TestAPI(unittest.TestCase):
+
+    def setUp(self):
+        self.app = app.test_client()
```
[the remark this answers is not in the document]
yashvardhannanavati: Yes, but then we will have to have a pytest global variable which will be used throughout all tests. I'll make the change

purview/base.py
yashvardhannanavati: I didn't quite get this one. Sorry!

tests/test_api_v1.py
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+import pytest
+
+from purview.api.v1 import app
+
+
+def pytest_namespace():
```
mprahl: Take a look at WaiverDB for how they handle the pytest fixture:
https://pagure.io/waiverdb/blob/master/f/tests/conftest.py

purview/api/v1.py
```diff
@@ -0,0 +1,22 @@
...
+def about():
+    """
+    Extract version info from setup.py and display it on /about endpoint
+    :return: version info
+    """
+    try:
+        version = pkg_resources.get_distribution('purview').version
+    except pkg_resources.DistributionNotFound:
+        version = 'unknown'
+    version_json = json.dumps({'Version': version})
```
mprahl: `Version` should be lowercase.

purview/__init__.py
```diff
@@ -3,4 +3,34 @@
 from __future__ import unicode_literals
 
 import logging
+
+from purview.error import (bad_request, not_found, ValidationError)
```
mprahl: Bike-shedding: You don't need parenthesis here

purview/api/v1.py
```diff
@@ -0,0 +1,22 @@
...
+
+    return version_json
```
mprahl: You need to use Flask's `jsonify` here.

purview/base.py
mprahl: Yes, create a base class that inherits from `StructuredNode` and then have our current models inherit from that base class. Then create the `serialize` function on that new base class. This will allow us to just call `serialize()` on any Neo4j node.

purview/base.py
```diff
@@ -0,0 +1,17 @@
...
+        """
+        Serializes a Node object to JSON
+        :param node_obj: a model object
+        :return: JSON string
+        """
+        return json.dumps(node_obj.__properties__, indent=4, sort_keys=True, default=str)
```
mprahl: This should return a dictionary instead of JSON so that way we can construct JSON with multiple nodes outside of this function.

purview/api/v1.py
```diff
@@ -0,0 +1,15 @@
...
+
+from flask import jsonify
+
+from purview import app, version
+
+
+@app.route('/about')
+def about():
+    """
+    Extract version info from setup.py and display it on /about endpoint
```
mprahl: Since the code changed, this function no longer extracts anything

purview/app.py
```diff
@@ -0,0 +1,44 @@
...
+from werkzeug.exceptions import default_exceptions
+
+from purview.logger import init_logging
+from purview.error import json_error, ValidationError
+from purview.api.v1 import api_v1
+
+
+def load_config(app):
+    default_config_file = None
+    if os.getenv('DEV') == 'true':
```
mprahl: This is fine but maybe we should do `if os.getenv('DEV') and os.getenv('DEV').lower() == 'true:` for ease.

mprahl:
:+1:

PR #34, opened 2018-04-04
dist-git scraper fixes

mprahl:
The first commit fixes a bug that occurs when the author or committer is in the wrong format in cgit. The second commit makes the cgit query faster and more efficient.


yashvardhannanavati:
Logic looks right to me. We can merge this after it's tested

PR #37, opened 2018-04-06
Add 'resource by id' endpoint

yashvardhannanavati:


purview/api/v1.py
```diff
@@ -17,3 +18,13 @@ def about():
...
+
+
+@api_v1.route('/<resource>/<id>')
+def get_resource(resource, id):
+    for model in all_models:
+        if model.__name__.lower() == resource:
+            try:
+                return model.nodes.get(id_=id).serialized
+            except model.DoesNotExist:
+                return 'Requested {0} does not exist!'.format(resource)
```
mprahl: This should be JSON. How about something like:
```python
@api_v1.route('/<resource>/<id>')
def get_resource(resource, id):
    for model in all_models:
        if model.__name__.lower() == resource:
            item = model.nodes.get_or_none(id_=id)
            if item:
                return jsonify(item.serialized)
    raise NotFound('The item does not exist.')
```


Then you'd need to do this import:
```python
from werkzeug.exceptions import NotFound
```

purview/api/v1.py
```diff
@@ -17,3 +18,13 @@ def about():
     :return: A JSON object with version info
     """
     return jsonify({'version': version})
+
+
+@api_v1.route('/<resource>/<id>')
+def get_resource(resource, id):
```
mprahl: I just realized `id` is overshadowing a built-in function. This should probably be renamed to `uid` or something similar.

purview/api/v1.py
```diff
@@ -17,3 +18,13 @@ def about():
...
+    for model in all_models:
+        if model.__name__.lower() == resource:
+            try:
+                return model.nodes.get(id_=id).serialized
```
mprahl: You need to actually find the `UniqueIdProperty` to filter against since it's different names on different nodes (e.g. `username`).

You can find the uid with:
```python
for prop_name, prop_def in model.defined_properties().items():
     if isinstance(prop_def, UniqueIdProperty):
         print('{0} is the uid!'.format(prop_name))
```

mprahl: The serialized output should be wrapped in a `jsonify` call.

yashvardhannanavati: Oh, you're right. Sorry about that!

yashvardhannanavati:
There's a bug. I'm still working on it. I'm trying to figure out if there's a way to replace the keyword in nodes.get() with a variable. Or else we might have to individually write it down for each model.

purview/api/v1.py
```diff
@@ -17,3 +20,15 @@ def about():
...
     return jsonify({'version': version})
+
+
+@api_v1.route('/<resource>/<uid>')
+def get_resource(resource, uid):
+    for model in all_models:
+        if model.__name__.lower() == resource:
+            for prop_name, prop_def in model.defined_properties().items():
+                if isinstance(prop_def, UniqueIdProperty):
+                    item = model.nodes.get(**{prop_def.name: uid})
```
mprahl: You should either change `model.nodes.get` to `model.nodes.get_or_none` or add an exception handler to Flask for `neomodel.DoesNotExist`.

purview/api/v1.py
```diff
@@ -17,3 +20,15 @@ def about():
...
+                    if item:
+                        return jsonify(item.serialized)
+    raise NotFound('This item does not exist')
```
mprahl: After thinking about this exception, I think should this should be in an else statement under `if item:` and then this line should have a different exception letting the user know they request an invalid resource. Something like:

```python
# TODO: This variable should be defined above and then the for loop could use this as well
model_names = [model.__name__.lower() for model in all_models]
error = 'The requested resource "{0}" is invalid. Choose from the following: {1}, and {2}.'.format(
    resource, ', '.join(model_names[:-1]), model_names[-1])
raise ValidationError()

File: purview/models/__init__.py [hunk 6]
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from purview.models.koji import KojiBuild, KojiTask, KojiTag
+from purview.models.bugzilla import BugzillaBug
+from purview.models.distgit import DistGitRepo, DistGitPush, DistGitBranch, DistGitCommit
+from purview.models.errata import Advisory, AdvisoryState
+from purview.models.freshmaker import FreshmakerEvent, ContainerBuilds
+from purview.models.user import User
+
+all_models = {KojiBuild, KojiTask, KojiTag, BugzillaBug, DistGitRepo, DistGitPush,
```
mprahl: Bike-shedding: Could this be alphabetized? That would make an error message to the user easier to read if we list out all the valid models.

mprahl:
@yashvardhannanavati done reviewing. Let me know if you have any questions.

yashvardhannanavati:
@mprahl Made the changes you suggested. Please let me know if this is correct

purview/api/v1.py
```diff
@@ -17,3 +21,20 @@ def about():
     :return: A JSON object with version info
     """
     return jsonify({'version': version})
+
+
+@api_v1.route('/<resource>/<uid>')
+def get_resource(resource, uid):
+    model_names = [model.__name__.lower() for model in all_models]
+    for model in all_models:
```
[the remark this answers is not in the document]
yashvardhannanavati: I missed this sorry. Accidently only changed it in my poc code.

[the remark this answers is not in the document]
yashvardhannanavati: Hence, I was thinking of keeping this as it is and only use the model_names for ValidationError

mprahl:
:+1: we just need unit tests on this. If you don't get around to it, I can write them later today.

yashvardhannanavati:
I just wrote a couple of comments above regarding how this change affects model.defined_properties().items()

yashvardhannanavati:
Change reverted to fix the bug

purview/api/v1.py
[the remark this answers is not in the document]
mprahl: As discussed in IRC, revert it back.

purview/api/v1.py
```diff
@@ -17,3 +21,20 @@ def about():
     :return: A JSON object with version info
     """
     return jsonify({'version': version})
+
+
+@api_v1.route('/<resource>/<uid>')
+def get_resource(resource, uid):
+    model_names = [model.__name__.lower() for model in all_models]
```
mprahl: This can actually be defined just before the error message. It saves the route from having to cycle through the list unnecessarily.

mprahl:
:ship:

PR #39, opened 2018-04-10
Add endpoint to give direct relationships of a certain resource

yashvardhannanavati:


purview/api/v1.py
```diff
@@ -23,15 +24,34 @@ def about():
     return jsonify({'version': version})
 
 
+def get_all_relationships(item):
```
mprahl: Can you move this to `utils.py`? Can you add a docstring?

purview/api/v1.py
```diff
@@ -23,15 +24,34 @@ def about():
...
+    all_relations = dict()
+    for model in all_models:
```
mprahl: Shouldn't we determine what models the item has a relationship to rather than cycle through all of them?

purview/api/v1.py
```diff
@@ -23,15 +24,34 @@ def about():
...
+        definition = dict(node_class=model, direction='OUTGOING')
+        relations_traversal = Traversal(item, model.__label__, definition)
+        all_model_relations = relations_traversal.all()
+        if len(all_model_relations) > 0:
+            all_relations[model.__name__] = all_model_relations
+    return all_model_relations
```
mprahl: Shouldn't this return `all_relations`?

purview/api/v1.py
```diff
@@ -23,15 +24,34 @@ def about():
...
+
+
 @api_v1.route('/<resource>/<uid>')
 def get_resource(resource, uid):
+    relationship = request.args.get('relationship', None)
```
mprahl: Default this to `False` instead of `None`.

purview/api/v1.py
```diff
@@ -23,15 +24,34 @@ def about():
...
     for model in all_models:
         if model.__name__.lower() == resource:
             for prop_name, prop_def in model.defined_properties().items():
                 if isinstance(prop_def, UniqueIdProperty):
                     item = model.nodes.get_or_none(**{prop_def.name: uid})
                     if item:
-                        return jsonify(item.serialized)
+                        if relationship:
```
mprahl: You need to actually parse the string to see if it's a boolean.

purview/api/v1.py
yashvardhannanavati: I was looking into this and an alternative approach would involve getting all relations of an item using item.__all_relationships__. Then traversing through the list of tuples returned and extracting unique model names. Then finding all nodes related to the item. Which I thought wouldn't be less computationally intense if not more. What are your thoughts on this?

purview/api/v1.py
yashvardhannanavati: Sure

yashvardhannanavati:
@mprahl Can you please review the changes?

purview/utils/general.py
```diff
@@ -33,3 +35,19 @@ def timestamp_to_datetime(timestamp):
...
+
+
+def get_all_relationships(item):
+    """
+    Gets all nodes having a relationship with the item node
+    :param item: a node (model) object
+    :return: a list of all nodes having relationship with item node
+    """
+    all_relations = {}
+    for model in all_models:
```
mprahl: As mentioned during our live review, this generates a query to Neo4j per model. So we'll need to refactor this to be more efficient.

PR #40, opened 2018-04-10
[WIP]Add global call to create app instance

yashvardhannanavati:


mprahl:
:+1: We'll need this when we deploy the app

PR #44, opened 2018-04-11
Fix Koji scraper bug

yashvardhannanavati:


scrapers/koji.py
```diff
@@ -91,7 +93,7 @@ def update_neo4j(self, builds):
             ))[0]
 
             user = User.get_or_create(dict(
-                username=build_dict['owner']
+                username=build_dict['owner_username']
```
mprahl: This is an email, so do this instead:
```python
build_dict['owner_username'].split('@')[0]
```

PR #48, opened 2018-04-11
Accept other Git URI formats from Koji build tasks

mprahl:
Some Git URIs for some reason only have `#` instead of `?#` between the repo and the commit hash.


yashvardhannanavati:
+1

PR #49, opened 2018-04-11
Accept more Git URI formats in the Koji scraper

mprahl:
Apparently some git URIs start with `git+https://` as well. If I encounter more errors like these, I'll convert this to a regex or be less aggressive in the filtering.


mprahl:
Okay, fixed!

scrapers/koji.py
```diff
@@ -128,7 +128,7 @@ def update_neo4j(self, builds):
                 xml_root = ET.fromstring(task_dict['request'])
                 commit_hash = None
                 for child in xml_root.iter('string'):
-                    if child.text.startswith('git://'):
+                    if child.text and child.text.startswith('git'):
```
yashvardhannanavati: This works too! (optional) The suggestion I made was having "if “git” in child.text" because that would guarantee that the url is detected no matter what it starts with. But if you are confident that it always starts with 'git' then we can keep this :smile:

PR #51, opened 2018-04-12
dist-git fixes and add parent/child relationships to commits

mprahl:


mprahl:
The scraper with these changes is now running and so far Neo4j looks good. @yashvardhannanavati, can you double-check and see if it looks right to you?

scrapers/distgit.py
```diff
@@ -143,6 +146,11 @@ def update_neo4j(self, results):
             commit.branches.connect(branch)
             commit.repos.connect(repo)
 
+            if repo_info['parent']:
+                parent_commit = DistGitCommit.get_or_create({'hash_': repo_info['parent']})[0]
+                commit.parents.connect(parent_commit)
+                parent_commit.children.connect(commit)
```
yashvardhannanavati: I thought yesterday we decided on not having child connections since they can be different for different branches. Did something change ?

scrapers/distgit.py
```diff
@@ -76,8 +76,11 @@ def update_neo4j(self, results):
                 'name': result['module']
             })[0]
             branch_name = result['ref'].rsplit('/', 1)[1]
-            branch = DistGitBranch.get_or_create(
-                {'name': branch_name}, relationships=repo.branches)[0]
+            branch = DistGitBranch.get_or_create({
+                'name': branch_name,
+                'repo_namespace': repo_info['namespace'],
```
yashvardhannanavati: Do you think having a relationship might still be better? I feel this is duplicating data and can be achieved through a relationship

scrapers/distgit.py
mprahl: Yes, after talking with @ralphbean we decided to do it since it's easy to implement with the current solution.

yashvardhannanavati: ah, I see :)

scrapers/distgit.py
mprahl: My original plan was to do it through a relationship but that ended up being rather confusing (having multiple nodes with the same values but different relationships). @ralphbean and I decided this approach is preferable even if it is data duplication.

scrapers/distgit.py
```diff
@@ -207,27 +215,45 @@ def _get_repo_info(self, repo_and_commit):
...
+                author = th_tag.next_sibling.string
+            elif th_tag.string == 'committer':
+                committer = th_tag.next_sibling.string
+            elif th_tag.string == 'parent':
+                rv['parent'] = th_tag.next_sibling.find('a').string
+
+            if author and committer and rv['parent']:
+                break
+
         for person in ('author', 'committer'):
```
mprahl: Instead of having a for loop, I could probably just turn this into a function and call it above. What do you think @yashvardhannanavati?

yashvardhannanavati: Yeap, I agree

mprahl:
This is ready for review again after doing some code cleanup.

yashvardhannanavati:
:+1:

PR #52, opened 2018-04-13
Style fixes and documentation

mprahl:
Review commit by commit for a rundown of the changes made

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,71 @@
-# Purview API
-An API for Purview
+# Purview API and Scrapers
 
+## Development
+
+To setup a development environment:
+* Create and activate a [Python virtual environment](https://virtualenv.pypa.io/en/stable/)
+    (Python 3 is preferred)
+* Install the API's dependencies with:
+  ```bash
+  $ pip install -r requirements.txt
+  ```
+* (Optional): Install the scrapers' dependencies with:
+  ```bash
+  $ pip install -r scraper-requirements.txt
+  ```
+
+To start the development web server, run:
+
+```bash
+$ scripts/run-flask.sh
+```
+
+
+## Run the Unit Tests
+
+Since the unit tests require a running Neo4j instance, the tests are run in Docker containers using
+Docker Compose. The commands required to run the unit tests are abstracted in
+`scripts/run-tests.sh`. This script will create the Docker image required to run the tests based
+on `docker/Dockerfile-tests`, create a container with `neo4j`, create another container to run
+the tests based on the built Docker image, and then delete the two created containers.
+
+To install Docker and Docker Compose on Fedora, run:
+
+```bash
+$ sudo dnf install docker docker-compose
+```
+
+To start Docker, run:
+
+```bash
+$ sudo systemctl start docker
+```
+
+To run the tests, run:
+
+```bash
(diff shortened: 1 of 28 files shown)


purview/api/v1.py
```diff
@@ -17,14 +17,25 @@
 @api_v1.route('/about')
 def about():
     """
-    Display version info about purview
-    :return: A JSON object with version info
+    Display general information about the app.
```
yashvardhannanavati: How about "...about the API" ?

purview/app.py
```diff
@@ -15,6 +15,11 @@
 
 
 def load_config(app):
+    """
+    Determine the correct configuration to use and apply it.
```
yashvardhannanavati: You can remove the "and apply it" part if you want. It seems like extra

purview/config.py
```diff
@@ -4,9 +4,8 @@
 
 
 class Config(object):
-    """
-    A Purview Flask configuration
-    """
+    """The base Purview application configuration."""
```
yashvardhannanavati: Any specific reason to have the quotes and the comment in a single line? If not then maybe we can have a uniform pattern throughout

README.md
```diff
@@ -1,3 +1,72 @@
 # Purview API
-An API for Purview
+The API and scrapers for Purview
```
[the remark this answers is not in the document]
mprahl: Sure

mprahl:
@yashvardhannanavati fixed. Is this good to merge now?

purview/models/distgit.py
```diff
@@ -38,6 +44,8 @@ class DistGitPush(PurviewStructuredNode):
 
 
 class DistGitCommit(PurviewStructuredNode):
+    """A representation of a dist-git commit in Neo4j."""
```
yashvardhannanavati: "Definition of a dist-git branch in Neo4j" ?

yashvardhannanavati: same for all other classes

purview/utils/general.py
```diff
@@ -10,9 +10,12 @@
 
 def timestamp_to_datetime(timestamp):
     """
-    Parses a timestamp and return a datetime object
-    :param timestamp: a string of a timestamp
-    :return: a datetime object
+    Convert a string timestamp to a datetime object.
+
+    :param str timestamp: a timestamp
```
yashvardhannanavati: "string in timestamp format"?

purview/api/v1.py
mprahl: Well we're actually exposing the app's version not the API so I think it's good as is.

purview/app.py
mprahl: I don't think it's implied by the function name so I'll keep it.

purview/config.py
mprahl: If you have a docstring that is just one line of text, it has to be one line like that based on the PEP guidelines.

purview/utils/general.py
mprahl: Changing to `:param str timestamp: a generic or ISO-8601 timestamp`

yashvardhannanavati:
:1st_place_medal:

PR #53, opened 2018-04-13
Add a test that tests the Bugzilla API with relationships

mprahl:
This also adds the ability to run just a single test


tests/api/test_bugzilla.py
```diff
@@ -0,0 +1,161 @@
...
+
+from purview.models.user import User
+from purview.models.bugzilla import BugzillaBug
+from purview.models.distgit import DistGitCommit
+from purview.models.errata import Advisory
+
+
+def test_get_bug(client):
+    """Test getting a resource from Neo4j with its relationships."""
+    tbrady = User.get_or_create({
```
yashvardhannanavati: Should we have general functions like 'create_user' and 'create bug' so that we can reduce duplication?

mprahl: I was thinking so but it doesn't save much code.

yashvardhannanavati: cool

yashvardhannanavati:
:+1:

PR #59, opened 2018-04-16
Add the inflate_node function

mprahl:


mprahl:
Seems like an issue with PyPi. The tests pass locally.

PR #64, opened 2018-04-17
Add code coverage using pytest plugin

yashvardhannanavati:


docker/Dockerfile-tests
```diff
@@ -16,11 +16,13 @@ RUN dnf -y install \
...
+  python3-pytest-cov \
   python3-tox \
   && dnf clean all
 RUN pip install flake8-docstrings
 
 VOLUME /src
 WORKDIR /src
 # Inspired from https://github.com/neo4j-contrib/neomodel/blob/master/tests-with-docker-compose.sh
-CMD ["sh", "-c", "while ! nc -z -w 2 neo4j 7687; do sleep 1; done; pytest-2 -vvv tests/ && pytest-3 -vvv tests/ && flake8"]
+CMD ["sh", "-c", "while ! nc -z -w 2 neo4j 7687; do sleep 1; done; pytest-2 -vvv tests/ && pytest-3 -vvv tests/ && pytest --cov=purview tests/ && flake8"]
```
mprahl: There's a way to run it as part of pytest:
https://pytest-cov.readthedocs.io/en/latest/readme.html#usage

mprahl: Here are some examples:
http://pytest-cov.readthedocs.io/en/latest/reporting.html

mprahl:
Here's an example of a config file I've used in the past so we can fail the coverage test if the coverage is below a certain percentage:
https://pagure.io/fedrepo_req/blob/master/f/.coveragerc

The initial percentage would be what's currently the result in the master branch.

mprahl:
:+1:

PR #70, opened 2018-04-18
Support node relationships to other nodes of the same type in serialized.all

mprahl:
@yashvardhannanavati can you verify if this fixes your bug? If so, I'll clean up the code.

### DIFF
--- a/purview/api/v1.py
+++ b/purview/api/v1.py
@@ -68,7 +68,7 @@ def get_resource_story(resource, uid):
     if not item:
         raise NotFound('This item does not exist')
 
-    for prop_def in item.defined_properties().values():
+    for _, prop_def in item.__all_properties__:
         if isinstance(prop_def, UniqueIdProperty):
             forward_query = create_query(item, prop_def.name, uid)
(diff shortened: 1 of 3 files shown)


yashvardhannanavati:
I tested this locally, the problem still persists :(

mprahl:
@yashvardhannanavati can you push your branch so I can test against it?

yashvardhannanavati:
Sure, actually the only change I made was remove the pdb and removing ".items()" from __all_properties__. Do you want me to overwrite on this branch?

yashvardhannanavati:
ah, my bad. I think you mean the tests I wrote. I'll do it

mprahl:
@yashvardhannanavati I fixed up the PR a bit. Try now.

mprahl:
So I tried running your tests with a few minor tweaks with this code and it seems to work except that Python3 returns `parent` as a list of dictionaries instead of a single dictionary. I'll investigate tomorrow.

mprahl:
@yashvardhannanavati cleaned up the PR. Could you review?

PR #73, opened 2018-04-19
Update bugzilla and errata scrapers to query on last_modified

yashvardhannanavati:
Update bugzilla and errata scrapers to query on last_modified rather than creation_date

### DIFF
--- a/scrapers/bugzilla.py
+++ b/scrapers/bugzilla.py
@@ -49,7 +49,7 @@ class BugzillaScraper(BaseScraper):
             LEFT JOIN bugzilla.profiles AS assigned ON bugs.assigned_to = assigned.userid
             LEFT JOIN bugzilla.profiles AS reported ON bugs.reporter = reported.userid
             LEFT JOIN bugzilla.profiles AS qa ON bugs.qa_contact = qa.userid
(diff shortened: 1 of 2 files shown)


scrapers/bugzilla.py
```diff
@@ -49,7 +49,7 @@ def get_bugzilla_bugs(self, start_date):
             LEFT JOIN bugzilla.profiles AS assigned ON bugs.assigned_to = assigned.userid
             LEFT JOIN bugzilla.profiles AS reported ON bugs.reporter = reported.userid
             LEFT JOIN bugzilla.profiles AS qa ON bugs.qa_contact = qa.userid
-            WHERE classifications.name = 'Red Hat' AND bugs.creation_ts > '{}'
+            WHERE classifications.name = 'Red Hat' AND bugs.lastdiffed > '{}'
```
mprahl: I think we actually want to use `delta_ts` here.

yashvardhannanavati: They are the same for all bugs :)

mprahl: Ah I see. This is the value that's exposed in the API, so I think we should use that one instead.

yashvardhannanavati: Agreed. Changed it :)

mprahl:
:+1:

PR #75, opened 2018-04-21
Add the "--days-ago" option the scraper script

mprahl:
This will be useful for when we run this on a cron job while we don't have the UMB listener/updater.

### DIFF
--- a/scripts/scrape.py
+++ b/scripts/scrape.py
@@ -17,7 +17,8 @@ log.setLevel(logging.DEBUG)
 
 parser = argparse.ArgumentParser(description='Run a scraper')
 parser.add_argument('scraper', type=str, help='The scraper to run. To run them all, use "all".')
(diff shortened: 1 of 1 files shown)


scripts/scrape.py
```diff
@@ -49,4 +57,7 @@
     scraper = scraper_class(
         args.teiid_user, args.teiid_password, args.kerberos, args.neo4j_user, args.neo4j_password,
         args.neo4j_server)
-    scraper.run(since=args.since)
+    since = args.since
+    if args.days_ago:
```
yashvardhannanavati: Maybe we should throw an error if the input given by the user is not an integer

mprahl: The `type=int` on line 23 should handle that already.

yashvardhannanavati:
:+1:

PR #79, opened 2018-04-23
Simplify the story API

mprahl:


mprahl:
@yashvardhannanavati fixed. Thanks for the review.

purview/api/v1.py
```diff
@@ -81,12 +80,15 @@ def get_resource_story(resource, uid):
...
-            results[curr_label] = results_unordered[curr_label]
+            # Have each index of results be a list with the 0 index being the resource name
+            # and the the rest of the indexes being resources
+            results_unordered[curr_label].insert(0, curr_label)
+            results.append(results_unordered[curr_label])
 
         curr_label = story_flow[curr_label]['forward_label']
 
-    return jsonify(list(results.items()) or item.serialized)
+    return jsonify(results)
```
yashvardhannanavati: Shouldn't this still have 'or item.serialized' ? Because we want to return the original item if we end up getting no relationships to it from neo4j

yashvardhannanavati:
@mprahl Fixed it!

tests/api/test_story.py
```diff
@@ -13,7 +13,7 @@
 
 
 def test_get_stories(client):
-    """Test getting a resource story from Neo4j with its relationships."""
+    '''Test getting a resource story from Neo4j with its relationships.'''
```
mprahl: This should stay a `"""`.

tests/api/test_story.py
```diff
@@ -221,8 +221,52 @@ def test_get_stories(client):
...
+    ]
+
+    rv = client.get('/api/v1/story/bugzillabug/5555')
+    assert rv.status_code == 200
+    assert json.loads(rv.data.decode('utf-8')) == expected
+
+
 def test_get_stories_not_available(client):
-    """Test getting a resource story on a resource that can't have a story."""
+    '''Test getting a resource story on a resource that can't have a story.'''
```
mprahl: Same as above.

mprahl:
@yashvardhannanavati just fix those quotes and lets merge. Thanks!

tests/api/test_story.py
yashvardhannanavati: Find and replace messed it up :P sorry!

PR #80, opened 2018-04-25
Fix bug when multiple resources of the same type are returned

yashvardhannanavati:


mprahl:
:+1: but let's add a test

mprahl:
:+1:

PR #82, opened 2018-04-27
Alter story endpoint to return an arbitary story for an artifact

yashvardhannanavati:


yashvardhannanavati:
@mprahl Nasty but works, phew! I intend to merge this only after the second feature we talked about is implemented. 

Also, we will have to install the 'apoc' plugin for the neo4j instance used in tests as well. :)

purview/utils/general.py
```diff
@@ -153,18 +138,21 @@ def create_query(item, uid_name, uid, reverse=False):
 
     while story_flow[curr_node_label][rel_label]:
         if curr_node_label == item.__label__:
-            node = node_query(curr_node_label, uid_name, uid)
-        else:
-            node = node_query(curr_node_label)
+            query += 'MATCH ({0}:{1} {{{2}:"{3}"}})\n'.format(
```
mprahl: It'd be easier to read if you did:
```python
query = """\
    MATCH ({var}:{label} {{{uid_name}:"{uid}"}})
    CALL apoc.path.expandConfig({var}, {{sequence:\'{label}
""".format(var=curr_node_label.lower(), label=curr_node_label, uid_name=uid_name.rstrip('_'), uid=uid)
```

mprahl: You could search by the Neo4j internal ID instead of by the uid. That would simplify the amount of arguments you need to pass and it's like 1ms faster too (not actually important haha)!

purview/api/v1.py
```diff
@@ -94,5 +94,8 @@ def get_resource_story(resource, uid):
     if len(results) == 0:
         # If no relationships are found, return the artifact by itself
         results.append([item.__label__, item.serialized])
+    else:
+        corelated_nodes = get_corelated_nodes(results, count=True)
```
mprahl: After discussion, we determined, `get_corelated_nodes` should accept `results_unordered` as the input instead of the formed JSON in `results`. This is because `results_unordered` is easier to work with and will not likely change where as the JSON being returned may change in the future.

purview/models/__init__.py
```diff
@@ -18,42 +18,53 @@
...
+            DistGitCommit.koji_builds.definition['relation_type']),
         'forward_label': KojiBuild.__label__,
-        'backward_relationship': DistGitCommit.resolved_bugs.definition['relation_type'],
+        'backward_relationship': '{0}>'.format(
+            DistGitCommit.resolved_bugs.definition['relation_type']),
         'backward_label': BugzillaBug.__label__
     },
     'KojiBuild': {
-        'forward_relationship': KojiBuild.advisories.definition['relation_type'],
+        'uid_name': 'id',
```
mprahl: To make this more dynamic, you could do:
```python
KojiBuild.id_.db_property or KojiBuild.id.name
```

purview/utils/general.py
```diff
@@ -151,20 +151,22 @@ def create_query(item, uid_name, uid, reverse=False):
...
-        next_node = node_query(story_flow[curr_node_label][node_label])
-        query += 'OPTIONAL MATCH {0}-[:{1}]-{2}\n'.format(
-            node, story_flow[curr_node_label][rel_label], next_node)
+        query += ', {0}, {1}'.format(
+            story_flow[curr_node_label][rel_label], story_flow[curr_node_label][node_label])
 
+        count += 1
         curr_node_label = story_flow[curr_node_label][node_label]
 
     if query:
```
mprahl: Optional: It'd be easier to read if this was `if count:`

purview/utils/general.py
```diff
@@ -151,20 +151,22 @@ def create_query(item, uid_name, uid, reverse=False):
...
-        query += 'RETURN *'
+        query += ('\', maxLevel: {0}}}) YIELD path\nRETURN path\n'
```
mprahl: As discussed, @yashvardhannanavati is being a good citizen and querying with explicit node labels and not `*`, so `maxLevel` is an unnecessary precaution.... we think.

mprahl: Optional: Make this a multi-line string for readability

purview/utils/general.py
```diff
@@ -183,6 +185,12 @@ def query_neo4j(query):
     if not results:
         return results_dict
 
+    if isinstance(results[0][0], int):
```
mprahl: Let's move this out to a `query_count_neo4j` function

purview/utils/general.py
```diff
@@ -183,6 +185,12 @@ def query_neo4j(query):
...
+        return results[0][0]-1
```
mprahl: When creating the separate function, don't forget to remove `-1` and have that done where the function is called so the separate function is generic.

purview/utils/general.py
```diff
@@ -183,6 +185,12 @@ def query_neo4j(query):
...
+
+    if isinstance(results[0][0], Path):
```
mprahl: Add a comment above saying that if a Path is in the first result, we assume that's all we want to process.

purview/utils/general.py
```diff
@@ -194,3 +202,38 @@ def query_neo4j(query):
                 if serialized_node not in results_dict[node_label]:
                     results_dict[node_label].append(inflated_node.serialized)
     return results_dict
+
+
+def get_corelated_nodes(results, count=True):
+    """
+    Create a raw cypher query for story nodes and get corelated nodes or their count.
```
mprahl: Add more to the docstring please :smile_cat:

purview/utils/general.py
```diff
@@ -194,3 +202,38 @@ def query_neo4j(query):
...
+    """
+    # To avoid circular imports
+    from purview.models import story_flow
+
+    queries_dict = {}
+    for artifact in results:
+        query = 'MATCH '
+        if story_flow[artifact[0]]['backward_label']:
+            node = node_query(artifact[0], story_flow[artifact[0]]['uid_name'], artifact[
```
mprahl: Can you make a variable of:
```python
uid_name = story_flow[artifact[0]]['uid_name']
```

purview/utils/general.py
```diff
@@ -110,7 +111,6 @@ def get_neo4j_node(resource_name, uid):
 def node_query(node_label, uid_name=None, uid=None):
```
mprahl: Can we rename this to `create_node_query`?

purview/utils/general.py
mprahl: Can we think of a better name here? To me it sounds like a resource is being returned?

purview/utils/general.py
```diff
@@ -194,3 +202,38 @@ def query_neo4j(query):
...
+    from purview.models import story_flow
+
+    queries_dict = {}
+    for artifact in results:
+        query = 'MATCH '
+        if story_flow[artifact[0]]['backward_label']:
+            node = node_query(artifact[0], story_flow[artifact[0]]['uid_name'], artifact[
+                              1][story_flow[artifact[0]]['uid_name']])
+            next_node = node_query(story_flow[artifact[0]]['backward_label'])
+            query += '{0}-[:{1}]-{2}\n'.format(node, story_flow[artifact[0]]
```
mprahl: Can you assign `story_flow[artifact[0]]['backward_relationship'][:-1]` to a variable for clarity?

mprahl:
Let's make this PR just about getting the count in correlated nodes, not the actual nodes itself. We'll figure out the rest in a later PR since this will be a separate PR.

purview/utils/general.py
yashvardhannanavati: We later discussed on removing the count, so I will keep this as it is :)

purview/utils/general.py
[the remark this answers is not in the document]
yashvardhannanavati: Need to think this through! I'll update the PR without this for the time-being

purview/api/v1.py
```diff
@@ -75,24 +76,28 @@ def get_resource_story(resource, uid):
 
     results_unordered = {}
     if forward_query:
+        print("Forward Query:", forward_query)
```
mprahl: This can be removed

purview/api/v1.py
```diff
@@ -75,24 +76,28 @@ def get_resource_story(resource, uid):
...
         results_unordered = query_neo4j(forward_query)
 
     if backward_query:
         results_unordered.update(query_neo4j(backward_query))
 
-    results = []
+    results = OrderedDict()
```
mprahl: Does this need to be an `OrderedDict`?

purview/api/v1.py
```diff
@@ -75,24 +76,28 @@ def get_resource_story(resource, uid):
...
     curr_label = 'BugzillaBug'
     while curr_label:
         if curr_label in results_unordered:
-            # Have each index of results be a list with the 0 index being the resource name
-            # and the the rest of the indexes being resources
-            results_unordered[curr_label].insert(0, curr_label)
-            results.append(results_unordered[curr_label])
+            if 'data' not in results:
```
mprahl: The results dictionary could just be initialized with this and you'd avoid this if statement.

purview/api/v1.py
```diff
@@ -75,24 +76,28 @@ def get_resource_story(resource, uid):
...
         curr_label = story_flow[curr_label]['forward_label']
 
-    if len(results) == 0:
+    if len(results['data']) == 0:
         # If no relationships are found, return the artifact by itself
-        results.append([item.__label__, item.serialized])
+        results['data'].append([item.__label__, item.serialized])
+
+    if len(results_unordered) > 0:
+        results['meta'] = {}
```
mprahl: The results dictionary could just be initialized with this.

purview/api/v1.py
```diff
@@ -75,24 +76,28 @@ def get_resource_story(resource, uid):
...
 
         curr_label = story_flow[curr_label]['forward_label']
 
-    if len(results) == 0:
+    if len(results['data']) == 0:
         # If no relationships are found, return the artifact by itself
-        results.append([item.__label__, item.serialized])
+        results['data'].append([item.__label__, item.serialized])
+
+    if len(results_unordered) > 0:
```
mprahl: This if statement should be removed because `get_corelated_nodes` should return a dictionary regardless if it doesn't already and the keys should be consistent.

purview/api/v1.py
```diff
@@ -75,24 +76,28 @@ def get_resource_story(resource, uid):
...
+                results['data'] = []
+            results['data'].append(results_unordered[curr_label][0])
 
         curr_label = story_flow[curr_label]['forward_label']
 
-    if len(results) == 0:
+    if len(results['data']) == 0:
         # If no relationships are found, return the artifact by itself
-        results.append([item.__label__, item.serialized])
+        results['data'].append([item.__label__, item.serialized])
```
mprahl: This seems like a different format than if the story is actual present.

purview/models/__init__.py
```diff
@@ -18,42 +18,53 @@
 story_flow = {
 
     'BugzillaBug': {
-        'forward_relationship': BugzillaBug.resolved_by_commits.definition['relation_type'],
+        'uid_name': BugzillaBug.id_.db_property or BugzillaBug.id.name,
```
mprahl: Optional: Getting the proper `uid_name` could be a separate function for less code repetition.

purview/utils/general.py
```diff
@@ -153,18 +154,21 @@ def create_query(item, uid_name, uid, reverse=False):
...
-        query += 'OPTIONAL MATCH {0}-[:{1}]-{2}\n'.format(
-            node, story_flow[curr_node_label][rel_label], next_node)
+        query += ', {0}, {1}'.format(
+            story_flow[curr_node_label][rel_label], story_flow[curr_node_label][node_label])
 
         curr_node_label = story_flow[curr_node_label][node_label]
 
     if query:
-        query += 'RETURN *'
+        query += ('\'}) YIELD path\nRETURN path\nORDER BY length(path) DESC\nLIMIT 1')
```
mprahl: Could you make this a multi-line string for readability please?

purview/api/v1.py
yashvardhannanavati: I kept this so that the order is always 'data' and then 'meta'.

purview/utils/general.py
```diff
@@ -191,6 +200,58 @@ def query_neo4j(query):
...
+
+    queries_dict = {}
+    for label, artifact in results.items():
+        query = 'MATCH '
+        backward_label = story_flow[label]['backward_label']
+        if backward_label:
+            backward_rel = story_flow[label]['backward_relationship'][:-1]
+            uid_name = story_flow[label]['uid_name']
+            node = create_node_subquery(label, uid_name,
+                                        artifact[0][uid_name])
```
mprahl: Bike-shedding: Can this be put on the previous line for readability?

purview/utils/general.py
```diff
@@ -191,6 +200,58 @@ def query_neo4j(query):
...
+            next_node = create_node_subquery(backward_label)
+            query += '{0}-[:{1}]-{2}\n'.format(node, backward_rel, next_node)
+            query += 'RETURN COUNT({0}) AS count'.format(
+                backward_label.lower())
+
+            queries_dict[backward_label] = query
+
+    nodes_count_dict = {}
+    for label, query in queries_dict.items():
```
mprahl: Rather than even having the `queries_dict`, I think you could just directly execute the Cypher in the previous loop (line 234 right now) and create `nodes_count_dict` as you go.

purview/api/v1.py
mprahl: Okay, it's not necessary but I don't have a problem with it either. You can keep it as is.

mprahl:
@yashvardhannanavati a few more comments but overall this looks great! Nice work!

purview/models/__init__.py
yashvardhannanavati: All I have is a string while getting the uid_name. Having a separate function would then involve looping through all_models everytime in order to get the model instance.

purview/utils/general.py
yashvardhannanavati: Nice idea, I'll make the change :)

purview/api/v1.py
```diff
@@ -4,12 +4,13 @@
 
 from flask import Blueprint, jsonify, request
 from werkzeug.exceptions import NotFound
+from neomodel import UniqueIdProperty
+from collections import OrderedDict
```
mprahl: This should be below line 3 because it's a standard import.

purview/api/v1.py
```diff
@@ -80,19 +81,18 @@ def get_resource_story(resource, uid):
...
-            results.append(results_unordered[curr_label])
+            results['data'].append(results_unordered[curr_label][0])
 
         curr_label = story_flow[curr_label]['forward_label']
 
-    if len(results) == 0:
+    if len(results['data']) == 0:
         # If no relationships are found, return the artifact by itself
-        results.append([item.__label__, item.serialized])
+        results['data'].append(item.serialized)
```
mprahl: This needs to have the `resource_type` key defined on it.

yashvardhannanavati: It already does

purview/models/__init__.py
mprahl: I meant something like
```python
def get_property_name(neomodel_property):
    return neomodel_property.db_property or neomodel_property.name

get_property_name(BugzillaBug.id_)
```

purview/utils/general.py
```diff
@@ -153,18 +154,27 @@ def create_query(item, uid_name, uid, reverse=False):
...
-        query += 'OPTIONAL MATCH {0}-[:{1}]-{2}\n'.format(
-            node, story_flow[curr_node_label][rel_label], next_node)
+        query += ', {0}, {1}'.format(
+            story_flow[curr_node_label][rel_label], story_flow[curr_node_label][node_label])
 
         curr_node_label = story_flow[curr_node_label][node_label]
 
     if query:
-        query += 'RETURN *'
+        # query += ('\'}) YIELD path\nRETURN path\nORDER BY length(path) DESC\nLIMIT 1')
```
mprahl: This can be deleted

purview/utils/general.py
```diff
@@ -191,6 +206,53 @@ def query_neo4j(query):
...
+        query = 'MATCH '
+        backward_label = story_flow[label]['backward_label']
+        if backward_label:
+            backward_rel = story_flow[label]['backward_relationship'][:-1]
+            uid_name = story_flow[label]['uid_name']
+            node_subquery = create_node_subquery(label, uid_name, artifact[0][uid_name])
+            next_node_subquery = create_node_subquery(backward_label)
+            query += '{0}-[:{1}]-{2}\n'.format(node_subquery, backward_rel, next_node_subquery)
+            query += 'RETURN COUNT({0}) AS count'.format(
+                backward_label.lower())
```
mprahl: Bike-shedding: Can this line be consolidated to the previous line?

purview/api/v1.py
mprahl: I don't think so since you add that key in `query_neo4j` and not `serialized`. That is an idea though, you could add the key directly in `serialized`.

tests/api/test_story.py
```diff
@@ -208,31 +200,27 @@ def test_get_stories(client):
...
-        'severity': 'unspecified',
-        'short_description': 'Fail to delete OSP tenant by CFME',
-        'status': 'CLOSED',
-        'target_milestone': 'GA',
-        'votes': 0
-    })
-
     rv = client.get('/api/v1/story/distgitcommit/8a63adb248ba633e200067e1ad6dc61931727bad')
     assert rv.status_code == 200
+    import pprint
```
mprahl: Can you remove this please?

purview/api/v1.py
```diff
@@ -80,19 +81,15 @@ def get_resource_story(resource, uid):
     if backward_query:
         results_unordered.update(query_neo4j(backward_query))
 
-    results = []
+    results = OrderedDict({'data': [], 'meta': {}})
+    results['meta']['related_nodes'] = {keys: 0 for keys in story_flow.keys()}
```
mprahl: Can you rename `keys` to `key` since it's not a list?

tests/api/test_story.py
```diff
@@ -282,14 +266,20 @@ def test_get_artifact_story_not_available(client):
...
                 'status': 'CLOSED',
                 'target_milestone': 'GA',
                 'votes': 0
             }
-        ]
-    ]
+        ],
+        'meta': {
+            'related_nodes': {
+
```
mprahl: This test will need to be fixed.

mprahl:
Nice! :tada:

PR #84, opened 2018-05-02
Remove required keyword on Freshmaker properties to avoid uniqueness violations

mprahl:
Due to how neomodel treats properties with `required=True` in `create_or_update` calls, this was causing some uniqueness violations.


mprahl:
@yashvardhannanavati it's because for whatever reason, neomodel uses all required properties to build the Cypher query in `get_or_create` or `create_or_update`, and so if one of those required properties changes, neomodel doesn't find the node and tries to create a new one.

It's really a bug in neomodel. It should only use unique properties but I haven't found the time to look into changing that.

PR #85, opened 2018-05-03
[WIP and Blocked] Add Product model to track releases

yashvardhannanavati:


yashvardhannanavati:
Once the structure looks good to go, I will fix the tests to suit the changes :)

purview/models/product.py
```diff
@@ -0,0 +1,18 @@
...
+
+from purview.models.base import PurviewStructuredNode
+
+
+class Product(PurviewStructuredNode):
+    """Definition of a released Product"""
+
+    attached_koji_tag = RelationshipFrom('.koji.KojiTag', 'ATTACHED_TO')
+    attached_advisory = RelationshipFrom('errata.Advisory', 'ATTACHED')
+    id_ = UniqueIdProperty(db_property='id')
```
mprahl: Is this unique ID coming from the Errata Tool? If so, we should consider moving this class to `errata.py`.

purview/models/product.py
```diff
@@ -0,0 +1,18 @@
...
+
+from neomodel import UniqueIdProperty, RelationshipFrom, StringProperty
+
+from purview.models.base import PurviewStructuredNode
+
+
+class Product(PurviewStructuredNode):
+    """Definition of a released Product"""
+
+    attached_koji_tag = RelationshipFrom('.koji.KojiTag', 'ATTACHED_TO')
```
mprahl: Attached aren't the right terms. How about `related` or `associated`? This should also be plural.

purview/models/product.py
```diff
@@ -0,0 +1,18 @@
...
+    attached_advisory = RelationshipFrom('errata.Advisory', 'ATTACHED')
```
mprahl: Same as above.

purview/models/koji.py
```diff
@@ -55,3 +55,4 @@ class KojiTag(PurviewStructuredNode):
     builds = RelationshipTo('KojiBuild', 'CONTAINS')
     id_ = UniqueIdProperty(db_property='id')
     name = StringProperty(required=True)
+    products = RelationshipTo('.product.Product', 'ATTACHED_TO')
```
mprahl: The `RelationshipTo` should likely be on the `Product` class because that is where the relationship is defined.

purview/models/errata.py
```diff
@@ -32,6 +31,7 @@ class Advisory(PurviewStructuredNode):
     assigned_to = RelationshipTo('.user.User', 'ASSIGNED_TO', cardinality=ZeroOrOne)
     attached_bugs = RelationshipTo('.bugzilla.BugzillaBug', 'ATTACHED')
     attached_builds = RelationshipTo('.koji.KojiBuild', 'ATTACHED')
+    attached_products = RelationshipTo('.product.Product', 'ATTACHED')
```
mprahl: I think this should just be called `product` and it should have cardinality of `ZeroOrOne` if I'm not mistaken.

purview/models/product.py
yashvardhannanavati: yes, but I made a separate file so that there is no confusion with the class. But sure, I can move it

purview/models/errata.py
yashvardhannanavati: I think it can be plural because one advisory can be attached to multiple products right.

purview/models/errata.py
```diff
@@ -33,6 +32,7 @@ class Advisory(PurviewStructuredNode):
     attached_bugs = RelationshipTo('.bugzilla.BugzillaBug', 'ATTACHED')
     attached_builds = RelationshipTo('.koji.KojiBuild', 'ATTACHED')
     package_owner = RelationshipTo('.user.User', 'PACKAGE_OWNED_BY', cardinality=ZeroOrOne)
+    products = RelationshipTo('Product', 'ASSOCIATED')
```
mprahl: @mikebonnet can an advisory have more than one product?

purview/models/errata.py
```diff
@@ -47,3 +47,14 @@ class AdvisoryState(PurviewStructuredNode):
     name = StringProperty(required=True)
     advisory = RelationshipTo('Advisory', 'STATE_OF', cardinality=ZeroOrOne)
     creator = RelationshipTo('.user.User', 'CREATED_BY', cardinality=ZeroOrOne)
+
+
+class Product(PurviewStructuredNode):
+    """Definition of a released Product"""
+
+    associated_koji_tag = RelationshipTo('.koji.KojiTag', 'ASSOCIATED_TO')
```
mprahl: This needs to be plural because there can be more than one Koji tag associated with an Errata product.

purview/models/errata.py
```diff
@@ -47,3 +47,14 @@ class AdvisoryState(PurviewStructuredNode):
...
+    associated_advisory = RelationshipFrom('Advisory', 'ASSOCIATED')
```
mprahl: This needs to be plural because more than one advisory can be associated with a product.

purview/models/errata.py
mprahl: The relationship name should be `ASSOCIATED` for consistency with the others.

yashvardhannanavati: Sure, we can change it to 'ASSOCIATED_WITH'

mprahl:
The model changes look great! Could you modify the scrapers now?

yashvardhannanavati:
Sure

scrapers/errata.py
```diff
@@ -198,3 +216,34 @@ def get_attached_bugs(self, advisory_id):
...
+        Query TEIID to find the associated koji tag from a product version ID.
+
+        :param int product_version_id: the product version ID
+        :return: a list of dictionaries
+        :rtype: list
+        """
+        sql = """\
+            SELECT tags.id AS tag_id, tags.name AS tag_name
+            FROM Errata_public.brew_tags_product_versions AS tags_pv,
+                 Errata_public.brew_tags AS tags
```
mprahl: Could you make this an explicit join instead of using the implicit syntax for readability?

scrapers/errata.py
```diff
@@ -69,6 +69,23 @@ def update_neo4j(self, advisories):
             reporter = User.get_or_create({'username': advisory['reporter'].split('@')[0]})[0]
             adv.conditional_connect(adv.reporter, reporter)
 
+            product_version_info = self.get_product_version_info(advisory['product_id'])
+            for item in product_version_info:
```
mprahl: Could you use a more descriptive name than item here?

scrapers/errata.py
```diff
@@ -198,3 +216,34 @@ def get_attached_bugs(self, advisory_id):
         """.format(advisory_id)
         log.info('Getting Bugzilla bugs tied to the advisory with ID {0}'.format(advisory_id))
         return self.teiid.query(sql)
+
+    def get_product_version_info(self, product_id):
```
mprahl: Could we merge this query into another join in `get_advisories`? If not, this is okay.

purview/models/errata.py
```diff
@@ -47,3 +47,14 @@ class AdvisoryState(PurviewStructuredNode):
     name = StringProperty(required=True)
     advisory = RelationshipTo('Advisory', 'STATE_OF', cardinality=ZeroOrOne)
     creator = RelationshipTo('.user.User', 'CREATED_BY', cardinality=ZeroOrOne)
+
+
+class Product(PurviewStructuredNode):
```
mprahl: After thinking about it, this should likely be called `ProductVersion`.

mprahl:
Some minor comments, but overall looks good. Let's test this once TEIID is back up.

scrapers/errata.py
yashvardhannanavati: Not directly, I think we might have to use nested queries to do that

purview/models/errata.py
yashvardhannanavati: hm, I am more inclined towards Product, since it's not explicitly for storing only the version_info.

mprahl:
This is blocked while we wait for an API from the ET.

mprahl:
Let's close this PR since it'll be quite a while until this is unblocked.

PR #86, opened 2018-05-04
Add an API end-point that lists the available resources and their UID

mprahl:
This will be used for the initial landing page on the frontend.

### DIFF
--- a/purview/api/v1.py
+++ b/purview/api/v1.py
@@ -52,6 +52,17 @@ def get_resource(resource, uid):
         return jsonify(item.serialized)
 
 
+@api_v1.route('/story')
+def get_available_resources():
+    """
+    List the available resources and their unique ID property.
+
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    """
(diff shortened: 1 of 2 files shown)


purview/api/v1.py
```diff
@@ -93,3 +93,19 @@ def get_resource_story(resource, uid):
...
+@api_v1.route('/resources')
+def get_resources():
+    """
+    List the available resources and their unique ID property.
+
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    """
+    resources = {}
+    for model in all_models:
```
yashvardhannanavati: It might be easier to iterate through the story_flow dict instead of all_models

yashvardhannanavati: since story_flow already has the uid of a resource as an attribute

mprahl:
@yashvardhannanavati fixed. Thanks for the review!

yashvardhannanavati:
There seems to be an unused import here. Can you please fix this

mprahl:
@yashvardhannanavati I noticed that a few mins ago and fixed it.

yashvardhannanavati:
:+1:

PR #87, opened 2018-05-07
Use serialized_all for the requested node on the story API endpoint

yashvardhannanavati:


yashvardhannanavati:
The code works but I think we will have to fix the tests

purview/utils/general.py
```diff
@@ -178,7 +178,7 @@ def create_query(item, uid_name, uid, reverse=False):
     return query
 
 
-def query_neo4j(query):
+def query_neo4j(query, resource):
```
mprahl: This function was written to be pretty generic and by adding this parameter it makes it specific to what the story API needs. How about making this `resources_to_expand` and have it be an optional list of labels the developer can provide that will have `serialized_all` be called on?

yashvardhannanavati: hm, how about I make an optional param which will have to be a list, and if it's present then '''python query_neo4j ''' will expand that node, else no.

yashvardhannanavati: Oh, I think we are telling the same thing. Sorry!

purview/utils/general.py
```diff
@@ -178,7 +178,7 @@ def create_query(item, uid_name, uid, reverse=False):
     return query
 
 
-def query_neo4j(query):
+def query_neo4j(query, resources_to_expand=None):
```
mprahl: At the beginning of the function, you should add:
```python
if resources_to_expand is None:
    resources_to_expand = []
```

mprahl:
:+1:

PR #90, opened 2018-05-11
Rename ContainerBuilds to ContainerBuild

mprahl:
We need this to happen before we deploy to prod.

Related: https://github.com/release-engineering/purview/pull/24


yashvardhannanavati:
:+1:

PR #93, opened 2018-05-16
Add all stories endpoint

yashvardhannanavati:


purview/utils/general.py
```diff
@@ -292,3 +291,76 @@ def get_node_count(query):
...
+    # list as the for loops will eliminate them. So we add the last element
+    # since we are sure it is unique.
+    unique_paths.append(results[0][0])
+
+    unique_paths_nodes = []
+    for path in unique_paths:
+        unique_paths_nodes.append(list(path.nodes))
+
+    print(unique_paths)
+    for index, result in enumerate(unique_paths_nodes):
```
mprahl: As we talked about, reuse the code from `get_story` by breaking it out into a function. `results_dict` should really be a list of dicts.

purview/utils/general.py
```diff
@@ -292,3 +291,76 @@ def get_node_count(query):
...
+    :rtype: dict
+    """
+    if resources_to_expand is None:
+        resources_to_expand = []
+
+    results_dict = {}
+    results, _ = db.cypher_query(query)
+
+    if not results:
+        return results_dict
```
mprahl: This needs to be `[results_dict]`

purview/utils/general.py
```diff
@@ -169,10 +169,9 @@ def create_query(item, uid_name, uid, reverse=False):
 
     if query:
         query += """\
-            \'}) YIELD path
+            \', minLevel:1}) YIELD path
             RETURN path
             ORDER BY length(path) DESC
-            LIMIT 1
```
mprahl: Make this a kwarg so that `get_story` doesn't get all the paths returned by Neo4j. It's just unnecessary churn.

purview/api/v1.py
```diff
@@ -104,3 +104,51 @@ def get_resource_story(resource, uid):
...
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    :raises NotFound: if the item is not found
+    :raises ValidationError: if an invalid resource was requested
+    """
+    item = get_neo4j_node(resource, uid)
+    if not item:
+        raise NotFound('This item does not exist')
+
+    for _, prop_def in item.__all_properties__:
```
mprahl: This could be a separate function since `get_resource_story` has the same code as well.

purview/utils/general.py
```diff
@@ -292,3 +291,76 @@ def get_node_count(query):
         return results_dict
 
     return results[0][0]
+
+
+def get_all_stories(query, resources_to_expand=None):
```
mprahl: Let's try consolidating this into the view function directly since it's not being reused anywhere.

purview/api/v1.py
```diff
@@ -103,4 +103,89 @@ def get_resource_story(resource, uid):
...
+        results_unordered_backward = []
+
+    all_results = []
+    try:
+        for result_forward in results_unordered_forward:
+            for result_backward in results_unordered_backward:
+                results_unordered = result_forward.copy()
+                results_unordered.update(result_backward)
+                results = OrderedDict({'data': [], 'meta': {}})
+                results['meta']['related_nodes'] = {key: 0 for key in story_flow.keys()}
```
mprahl: You should assign `{key: 0 for key in story_flow.keys()}` to a variable above the for loop. That way you don't recreate this every time.

purview/api/v1.py
```diff
@@ -103,4 +103,89 @@ def get_resource_story(resource, uid):
...
+    else:
+        results_unordered_backward = []
+
+    all_results = []
+    try:
+        for result_forward in results_unordered_forward:
+            for result_backward in results_unordered_backward:
+                results_unordered = result_forward.copy()
+                results_unordered.update(result_backward)
+                results = OrderedDict({'data': [], 'meta': {}})
```
mprahl: This doesn't need to be an OrderedDict.

purview/api/v1.py
```diff
@@ -103,4 +103,89 @@ def get_resource_story(resource, uid):
...
+                results['meta']['related_nodes'] = {key: 0 for key in story_flow.keys()}
+                curr_label = 'BugzillaBug'
+                while curr_label:
```
mprahl: Are the results from Neo4j not returned in order? If not, could you make this a separate function?

purview/api/v1.py
```diff
@@ -103,4 +103,89 @@ def get_resource_story(resource, uid):
...
+                    if curr_label in results_unordered:
+                        results['data'].append(results_unordered[curr_label][0])
+
+                    curr_label = story_flow[curr_label]['forward_label']
+
+                results['meta']['related_nodes'].update(get_corelated_nodes(results_unordered))
+            all_results.append(results)
+
+    except UnboundLocalError:
```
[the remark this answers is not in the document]
mprahl: I don't understand why you have this except, but it's hard to read and likely not what we want to do considering it's repeating a lot of code below.

purview/api/v1.py
```diff
@@ -103,4 +103,89 @@ def get_resource_story(resource, uid):
...
+            while curr_label:
+                if curr_label in result:
+                    results['data'].append(result[curr_label][0])
+
+                curr_label = story_flow[curr_label]['forward_label']
+
+            results['meta']['related_nodes'].update(get_corelated_nodes(result))
+        all_results.append(results)
+
+    # Adding the artifact itself if it's story is not available
```
mprahl: "it's" => "its"

purview/api/v1.py
```diff
@@ -103,4 +103,89 @@ def get_resource_story(resource, uid):
...
+    if len(all_results) == 0:
+        results = OrderedDict({'data': [], 'meta': {}})
```
mprahl: This doesn't need to be an OrderedDict.

purview/utils/general.py
```diff
@@ -169,12 +169,14 @@ def create_query(item, uid_name, uid, reverse=False):
...
         query += """\
-            \'}) YIELD path
+            \', minLevel:1}) YIELD path
             RETURN path
             ORDER BY length(path) DESC
-            LIMIT 1
             """
 
+    if query and limit:
+        query += """ LIMIT 1"""
```
mprahl: Change `""" LIMIT 1"""` to `' LIMIT 1'`.

purview/utils/general.py
```diff
@@ -186,37 +188,21 @@ def query_neo4j(query, resources_to_expand=None):
     :return results_dict: a dictionary containing serialized results received from neo4j
```
mprahl: This needs updating

purview/utils/general.py
```diff
@@ -292,3 +278,88 @@ def get_node_count(query):
...
+        return [results_list]
+
+    # Creating a list of lists where each list is a collection of node ids
+    # of the nodes present in that particular story path.
+    # Paths are re-sorted in the ascending order for simplifying the logic below
+    path_nodes_id = []
+    for path in reversed(results):
+        path_nodes_id.append([node.id for node in path[0].nodes])
+
+    print("path nodes:", path_nodes_id)
```
mprahl: This can be removed

purview/utils/general.py
```diff
@@ -292,3 +278,88 @@ def get_node_count(query):
...
+    unique_paths = []
+    for index, node_set in enumerate(path_nodes_id[:-1]):
+        unique = True
+        for alternate_set in path_nodes_id[index + 1:]:
+            # If the node_set is a subset of alternate_set,
+            # we know they are the same path except the alternate_set is longer.
+            # If alternate_set and node_set only have one node_id of difference,
+            # we know it's the same path but with the perspective of different siblings.
```
mprahl: "with the" => "from the"

purview/utils/general.py
```diff
@@ -292,3 +278,88 @@ def get_node_count(query):
...
+            if set(node_set).issubset(set(alternate_set)) or len(
+                    set(alternate_set).difference(set(node_set))) == 1:
+                unique = False
+                break
+        if unique:
+            # Since results is from longest to shortest, we need to get
+            # the opposite index.
+            unique_paths.append(results[(len(path_nodes_id) - index) - 1][0])
+    # While traversing, the outer for loop only goes until the second-last element
```
mprahl: "second-last" => "second to last"

purview/api/v1.py
yashvardhannanavati: No, I don't think they are returned in order. Sure, I will make it a separate function.

purview/utils/general.py
yashvardhannanavati: The view function will then grow very large, do we still want to do it?

purview/api/v1.py
yashvardhannanavati: Can you suggest a nice name for this? since we already have a `create_query` function.

mprahl:
Looks good to me. Just fix the import issue and add unit tests.

purview/api/v1.py
```diff
@@ -79,18 +82,34 @@ def get_resource_story(resource, uid):
...
-        if isinstance(prop_def, UniqueIdProperty):
-            forward_query = create_query(item, prop_def.name, uid)
-            backward_query = create_query(item, prop_def.name, uid, reverse=True)
-            break
+    uid_name = item.unique_id_property
+    forward_query = create_story_query(item, uid_name, uid, limit=True)
+    backward_query = create_story_query(item, uid_name, uid, reverse=True,
+                                        limit=True)
+
+    def _get_partial_story(query, resources_to_expand=None):
```
mprahl: We can remove the `=None` part now since we know this function will always be called with an appropriate value now.

purview/api/v1.py
```diff
@@ -79,18 +82,34 @@ def get_resource_story(resource, uid):
...
+        results_list = []
+        results, _ = db.cypher_query(query)
+
+        if resources_to_expand is None:
```
mprahl: This if statement can be removed after following my comment above.

purview/api/v1.py
```diff
@@ -79,18 +82,34 @@ def get_resource_story(resource, uid):
...
+            resources_to_expand = []
+
+        if not results:
+            return [results_list]
+
+        # Assuming that if Path is the first result, then that's all we want to process.
+        if isinstance(results[0][0], Path):
```
mprahl: This will always be a path now when results are returned, right? Perhaps this if statement can be removed and just always execute the code below.

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
 
     results['meta']['related_nodes'].update(get_corelated_nodes(results_unordered))
 
+    # Adding the artifact itself if it's story is not available
+    if len(results['data']) == 0:
```
mprahl: Optional: This check could be moved up in the function (with some changes) so that the while loop is not needlessly executed when there is no path.

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+    """
+    item = get_neo4j_node(resource, uid)
+    if not item:
+        raise NotFound('This item does not exist')
+
+    uid_name = item.unique_id_property
+    forward_query = create_story_query(item, uid_name, uid)
+    backward_query = create_story_query(item, uid_name, uid, reverse=True)
+
+    def _get_partial_stories(query, resources_to_expand=None):
```
mprahl: Same comment as in get_resource_story.

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+        if resources_to_expand is None:
```
mprahl: Same comment as in get_resource_story.

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+            resources_to_expand = []
+
+        results_list = []
+        results, _ = db.cypher_query(query)
+
+        if not results:
+            return [results_list]
+
+        # Creating a list of lists where each list is a collection of node ids
```
mprahl: Optional: "ids" => "IDs"

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+        # of the nodes present in that particular story path.
+        # Paths are re-sorted in the ascending order for simplifying the logic below
```
mprahl: "in the ascending" => "in ascending"

mprahl: "for simplifying" => "to simplify"

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+        path_nodes_id = []
+        for path in reversed(results):
+            path_nodes_id.append([node.id for node in path[0].nodes])
+
+        unique_paths = []
+        for index, node_set in enumerate(path_nodes_id[:-1]):
+            unique = True
+            for alternate_set in path_nodes_id[index + 1:]:
+                # If the node_set is a subset of alternate_set,
```
mprahl: Optional: You could take some of the words from the line below and put it at the end of this line. It's a waste of space have the line break so early.

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+                # we know they are the same path except the alternate_set is longer.
+                # If alternate_set and node_set only have one node_id of difference,
```
mprahl: "node_id" => "node ID" since "node_id" is not a variable and that format implies that it is one.

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+    else:
+        results_unordered_forward = []
+
+    if backward_query:
+        results_unordered_backward = _get_partial_stories(backward_query, [resource])
+    else:
+        results_unordered_backward = []
+
+    all_results = []
+    if len(results_unordered_backward) == 0 or len(results_unordered_forward) == 0:
```
mprahl: Optional: You could just do:
`if not results_unordered_backward or not results_unordered_forward:`

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+        if len(results_unordered_forward) > 0:
```
mprahl: Optional: Same as above

purview/api/v1.py
```diff
@@ -103,4 +122,110 @@ def get_resource_story(resource, uid):
...
+            results_unordered_unidir = results_unordered_forward
+        else:
+            results_unordered_unidir = results_unordered_backward
+
+        for result in results_unordered_unidir:
+            all_results.append(_order_story_results(result))
+
+    else:
+        for result_forward in results_unordered_forward:
```
mprahl: Above this, could you please add a comment like:
```python
# Combining all the backward and forward paths to generate all the possible full paths
```

purview/utils/general.py
```diff
@@ -279,10 +239,10 @@ def get_corelated_nodes(results):
 
 def get_node_count(query):
     """
-    Query neo4j and return the count of results.
+    Query Neo4j and return the count of results.
 
     :param str query: raw cypher query
-    :return results_dict: a dictionary containing count of results received from neo4j.
+    :return: a dictionary containing count of results received from Neo4j.
```
mprahl: "containing count of results" => "containing the results count"

purview/api/v1.py
mprahl: It works for me:
```python
>>> results_unordered_backward = []
>>> results_unordered_forward = [1, 2, 3]
>>> if not results_unordered_backward or not results_unordered_forward:
...     print('hello')
... 
hello
```

yashvardhannanavati: oops! looks like I had a typo.

mprahl:
:+1: this can be merged after the tests pass

PR #94, opened 2018-05-17
Update the scrapers OpenShift template

mprahl:
This PR:
* Drops the out of date and unused Neo4j OpenShift template
* Adds the `SCRAPE_FROM_DAYS_AGO` environment variable to the scraper Docker image so that the scraper start dates can be customized
* Splits out the the secret creation from the scraper OpenShift template to a separate template since secret creation only needs to happen once while you may want to run the scraper job multiple times
* Use the new scraper Docker image built by Jenkins and published to Quay.io instead of building it in OpenShift directly from source


mprahl:
I just realized this should be in a separate repo. I'll refactor the PR to just include the Dockerfile changes.

PR #100, opened 2018-05-18
Install Purview in the Docker image

mprahl:
Install Purview in the Docker image so that the about API endpoint shows the version.


docker/Dockerfile-api
```diff
@@ -15,5 +15,6 @@ RUN dnf -y install \
 # This will allow a non-root user to install a custom root CA at run-time
 RUN chmod 777 /etc/pki/tls/certs/ca-bundle.crt
 COPY . .
+RUN python3 setup.py install --prefix /usr
```
mprahl: The default prefix is something like `/usr/local` which leads to a Python 3 path that doesn't exist and thus setup.py fails. Rather than create the folder, I chose to install it where the Python 3 modules packaged as RPMs install to.

PR #102, opened 2018-05-21
Add the ability to keep trying Teiid when it is down with a backoff

mprahl:


purview/utils/teiid.py
```diff
@@ -56,35 +58,44 @@ def get_connection(self, db_name):
...
-                if attempts < retry:
-                    log.warning('Teiid query failed')
-                else:
+                if retry is not None and attempts > retry:
                     raise
+                else:
+                    # Double the backoff time
+                    backoff = backoff * 2
+                    # Max out the backoff time to 15 minutes
+                    backoff = min((backoff, fifteen_mins))
```
yashvardhannanavati: Instead of doing this, I would suggest checking `if backoff < fifteen_mins:` before doing `'backoff*2'` because the current approach might end up taking a lot of memory in case teiid goes down for a day or so

mprahl: Good point. I fixed that.

purview/utils/teiid.py
```diff
@@ -56,35 +59,46 @@ def get_connection(self, db_name):
...
+        :kwarg int retry: the  number of times to retry a failed query. If this
+        is not set, then the TEIID query will be repeated until it is successful.
         :return: a list of rows from Teiid. Each row is a dictionary
         with the column headers as the keys.
         :rtype: list
         """
         con = self.get_connection(db)
         cursor = con.cursor()
-        if retry < 1:
+        if retry is not None and retry < 1:
```
yashvardhannanavati: Can we do `if retry and retry < 1:`?

mprahl: That should be fine.

PR #103, opened 2018-05-21
Force a new TEIID connection on query failures

mprahl:


purview/utils/teiid.py
```diff
@@ -99,6 +101,8 @@ def query(self, sql, db='public', retry=None):
                     log.warning('The Teiid query failed on attempt {0}. Sleeping for {1} seconds.'
                                 .format(attempts, backoff))
                     sleep(backoff)
+                    # Restart the database connection
+                    con = self.get_connection(db, force_new=True)
```
mprahl: I just realized this should be in the try part above. I'll fix this.

mprahl: Fixed

PR #104, opened 2018-05-22
Retry on Teiid connect and fix retry on query

mprahl:


purview/utils/teiid.py
```diff
@@ -29,29 +29,44 @@ def __init__(self, host, port, username, password):
...
         :kwarg bool force_new: forces a new database connection even if one
         already exists
+        :kwarg int retry: the number of times to retry a failed connection. If this
+        is not set, then the TEIID connection attempt will be repeated until it is successful.
         :return: a connection to TEIID
         :rtype: psycopg2 connection
         """
         if not force_new and db_name in self._connections:
             return self._connections[db_name]
+        if retry is not None and retry < 1:
```
yashvardhannanavati: Optional: We can change this to `if retry and retry < 1:`

yashvardhannanavati:
:+1:

purview/utils/teiid.py
mprahl: The reason I use `if retry is not None` is because I want this exception to be raised if `retry` is set to `0`.

PR #105, opened 2018-05-22
Move teiid.py and fix capitalization of Teiid in comments and log messages

mprahl:
teiid.py is only used in the scrapers so it seems more appropriate for it to be in the scrapers folder. We also haven't written unit tests for the scrapers since they are meant to be one-time scripts and then replaced with a message bus listener, so it brings down our coverage report.


yashvardhannanavati:
:+1:

PR #110, opened 2018-05-24
Show Container builds as Brew builds scrapers and model fixes

yashvardhannanavati:


scrapers/freshmaker.py
```diff
@@ -61,34 +63,40 @@ def query_api_and_update_neo4j(self):
...
-                        type_=build['type'],
-                        type_name=build['type_name'],
-                        url=build['url']
-                    )
-                    if build['time_completed']:
-                        cb_params['time_completed'] = timestamp_to_datetime(
-                            build['time_completed'])
-                    cb = ContainerBuild.create_or_update(cb_params)[0]
-                    event.triggered_container_builds.connect(cb)
+                    # The build ID obtained from Freshmaker API is actually a koji task ID
```
mprahl: Optional: "from Freshmaker" => "from the Freshmaker"
"koji" => "Koji"

scrapers/freshmaker.py
```diff
@@ -61,34 +63,40 @@ def query_api_and_update_neo4j(self):
...
+                    task_result = self.get_task_results(build_dict['build_id'])[0]
+
+                    # Extract the build ID from a task result
+                    xml_root = ET.fromstring(task_result['result'])
+                    build_id = xml_root.find(".//*[name='koji_builds'].//string")
+
+                    build = KojiBuild.get_or_create(dict(id_=build_id))[0]
```
mprahl: What if `build_id` is `None`?

scrapers/freshmaker.py
```diff
@@ -2,11 +2,13 @@
 
 from __future__ import unicode_literals
 
+import xml.etree.ElementTree as ET
```
mprahl: Optional: This can be moved up one line

scrapers/freshmaker.py
```diff
@@ -61,34 +63,43 @@ def query_api_and_update_neo4j(self):
...
-                        url=build['url']
-                    )
-                    if build['time_completed']:
-                        cb_params['time_completed'] = timestamp_to_datetime(
-                            build['time_completed'])
-                    cb = ContainerBuild.create_or_update(cb_params)[0]
-                    event.triggered_container_builds.connect(cb)
+                    # Extract the build ID from a task result
+                    xml_root = ET.fromstring(task_result['result'])
+                    build_id = xml_root.find(".//*[name='koji_builds'].//string")
```
mprahl: It worries me that `koji_builds` is plural. Have you seen one in Teiid that has more than one build? It could just be a grammar error by the developer.

yashvardhannanavati: hm, that's an interesting observation. Maybe we should confirm it with someone

yashvardhannanavati: I have added a TODO there and I will also create an issue, if we happen to find out that there can be multiple builds then we can change it. For the time being, I am merging this so that it doesn't block the API end-point change.

yashvardhannanavati: Asked @ralphbean on IRC about this and he thinks that a task can only trigger one build. Hence, not creating an issue for this.

PR #111, opened 2018-05-24
Fix bug to include container build tasks

yashvardhannanavati:


scrapers/koji.py
```diff
@@ -147,6 +148,11 @@ def update_neo4j(self, builds):
                 release=build_dict['release']
             ))[0]
 
+            try:
+                extra_json = json.loads(build_dict['extra'])
```
mprahl: Since this variable isn't used until 30 or so lines later, could you move the code down to where it's used for readability?

PR #114, opened 2018-05-29
Fix the story API when the artifact doesn't have any forward relationships

mprahl:


yashvardhannanavati:
It looks like there is an unused local variable. Can you please remove it and fix the build?

yashvardhannanavati:
It's good to go after that

PR #115, opened 2018-05-30
Fix API end-points to show container builds as Koji builds

yashvardhannanavati:


yashvardhannanavati:
I have modified the API endpoints to suit our scraper and model changes. The test cases will fail because they still expect container builds as a part of FreshmakerEvent and not KojiBuild

purview/api/v1.py
```diff
@@ -105,26 +105,15 @@ def _get_partial_story(query, resources_to_expand):
         results_unordered.update(
             _get_partial_story(backward_query, [resource]))
 
-    results = OrderedDict({'data': [], 'meta': {}})
-    results['meta']['related_nodes'] = {key: 0 for key in story_flow.keys()}
-
     # Adding the artifact itself if it's story is not available
     if len(results_unordered) == 0:
+        results = OrderedDict({'data': [], 'meta': {}})
```
mprahl: This doesn't need to be an OrderedDict

purview/api/v1.py
```diff
@@ -184,7 +173,6 @@ def _get_partial_stories(query, resources_to_expand):
         # since we are sure it is unique.
         unique_paths.append(results[0][0])
         unique_paths_nodes = [path.nodes for path in unique_paths]
-
```
mprahl: This is an unnecessary edit.

purview/models/__init__.py
```diff
@@ -81,3 +37,62 @@ def replace(self, node, properties=None):   # pragma: no cover
 
 RelationshipManager.disconnect_all = disconnect_all
 RelationshipManager.replace = replace
+
+
+def story_flow(label):
```
mprahl: Needs a docstring

mprahl: The variable name `label` is misleading because I assume it to mean it's a Neo4j label which is now no longer true (e.g. `ContainerKojiBuild`.)

purview/models/__init__.py
```diff
@@ -81,3 +37,62 @@ def replace(self, node, properties=None):   # pragma: no cover
...
+        }
+    elif label == 'ContainerKojiBuild':
+        return {
+            'uid_name': KojiBuild.id_.db_property or KojiBuild.id.name,
+            'forward_relationship': None,
+            'forward_label': None,
+            'backward_relationship': '{0}<'.format(
+                KojiBuild.triggered_by_freshmaker_event.definition['relation_type']),
+            'backward_label': FreshmakerEvent.__label__
+        }
```
mprahl: You should add an else that raises an exception here.

purview/models/__init__.py
```diff
@@ -81,3 +37,62 @@ def replace(self, node, properties=None):   # pragma: no cover
...
+            'backward_relationship': '{0}>'.format(
+                Advisory.attached_builds.definition['relation_type']),
+            'backward_label': KojiBuild.__label__
+        }
+    elif label == 'FreshmakerEvent':
+        return {
+            'uid_name': FreshmakerEvent.id_.db_property or FreshmakerEvent.id.name,
+            'forward_relationship': '{0}>'.format(
+                FreshmakerEvent.triggered_container_builds.definition['relation_type']),
+            'forward_label': 'Container'+KojiBuild.__label__,
```
mprahl: Please use `.format` for consistency in string manipulation.

purview/models/__init__.py
```diff
@@ -81,3 +37,62 @@ def replace(self, node, properties=None):   # pragma: no cover
...
+            'backward_relationship': '{0}>'.format(FreshmakerEvent.triggered_by_advisory
+                                                   .definition['relation_type']),
+            'backward_label': Advisory.__label__
+        }
+    elif label == 'ContainerKojiBuild':
```
mprahl: Could you make this like the following for consistency:
`elif label == 'Container{0}'.format(KojiBuild.__label__):`

purview/models/base.py
```diff
@@ -183,7 +183,17 @@ def inflate_results(results, resources_to_expand):
                     else:
                         serialized_node = inflated_node.serialized
 
-                    serialized_node['resource_type'] = node_label
+                    if node_label == 'KojiBuild' and 'type_' in serialized_node:
```
mprahl: This whole if statement could be reduced by overriding the `node_label` variable above when it is a `KojiBuild` of type `container`.

yashvardhannanavati:
@mprahl I have made all the changes that we had discussed. Can you please review this?

purview/models/__init__.py
```diff
@@ -81,3 +37,71 @@ def replace(self, node, properties=None):   # pragma: no cover
 
 RelationshipManager.disconnect_all = disconnect_all
 RelationshipManager.replace = replace
+
+
+def story_flow(label):
+    """
+        Get the next/previous node in a story flow/pipeline path
```
mprahl: The docstring contents are indented one too many times. This also needs a period at the end of it.

purview/models/base.py
```diff
@@ -184,6 +184,7 @@ def inflate_results(results, resources_to_expand):
                         serialized_node = inflated_node.serialized
 
                     serialized_node['resource_type'] = node_label
+
```
mprahl: Could you remove this errant new line please?

purview/utils/general.py
```diff
@@ -92,6 +92,9 @@ def get_neo4j_node(resource_name, uid):
     # To prevent a ciruclar import, we must import this here
     from purview.models import all_models
 
+    if resource_name.lower() == 'kojicontainerbuild':
```
mprahl: Could you please delete this if statement and add `KojiContainerBuild` to `all_models`. That should remove the need for this.

mprahl:
@yashvardhannanavati left a few comments.

yashvardhannanavati:
@mprahl Fixed it! :)

purview/api/v1.py
```diff
@@ -105,26 +105,15 @@ def _get_partial_story(query, resources_to_expand):
         results_unordered.update(
             _get_partial_story(backward_query, [resource]))
 
-    results = OrderedDict({'data': [], 'meta': {}})
-    results['meta']['related_nodes'] = {key: 0 for key in story_flow.keys()}
-
     # Adding the artifact itself if it's story is not available
     if len(results_unordered) == 0:
+        results = {'data': [], 'meta': {}}
```
mprahl: Optional: You could call `_order_story_results` instead of these two lines.

purview/models/__init__.py
```diff
@@ -81,3 +37,71 @@ def replace(self, node, properties=None):   # pragma: no cover
...
+
+def story_flow(label):
+    """
+    Get the next/previous node in a story flow/pipeline path.
+
+    :param str label: Neo4j node label
+    :return: uid and relationship information in both forward and backward directions
+    :rtype: dict
+    """
+    if label == 'BugzillaBug':
```
mprahl: Optional: I'd prefer if we used `BugzillaBug.__label__` and the same pattern below

purview/models/__init__.py
```diff
@@ -81,3 +37,71 @@ def replace(self, node, properties=None):   # pragma: no cover
 
 RelationshipManager.disconnect_all = disconnect_all
 RelationshipManager.replace = replace
+
+
+def story_flow(label):
```
mprahl: This should likely be moved to `utils.py`

purview/api/v1.py
yashvardhannanavati: That would just become confusing, hence I would like it to be as it is. :)

PR #117, opened 2018-05-30
Add another label for KojiBuild model

yashvardhannanavati:


purview/models/koji.py
```diff
@@ -32,6 +32,11 @@ class KojiBuild(PurviewStructuredNode):
     version = StringProperty()
 
 
+class ContainerKojiBuild(KojiBuild):
```
mprahl: How about `KojiContainerBuild`?

yashvardhannanavati: Haha, I had this dilemma yesterday as well, but then I thought that since we have a model class called KojiBuild, naming it ContainerKojiBuild might just give a hint of connection between the two classes to someone who just sees the labels. I know it sounds stupid, but what do you think?

mprahl: Sure, either one works for me.

purview/models/koji.py
```diff
@@ -32,6 +32,11 @@ class KojiBuild(PurviewStructuredNode):
     version = StringProperty()
```
mprahl: Could you please remove `type_ = StringProperty()` above this?

PR #119, opened 2018-06-01
Add  argument to support date range for scraping

yashvardhannanavati:


scrapers/base.py
```diff
@@ -13,8 +13,9 @@ class BaseScraper(object):
 
     teiid_host = 'virtualdb.engineering.redhat.com'
     teiid_port = 5432
-    # Default start date to fetch data from
+    # Default start date and end date to fetch data
     default_since = datetime(2016, 3, 1)
```
mprahl: Optional: I know you didn't set this in this PR but it'd be nice to default to 365 days ago instead of a static date.

scrapers/base.py
```diff
@@ -13,8 +13,9 @@ class BaseScraper(object):
...
+    default_until = datetime.today().strftime('%Y-%m-%d')
```
mprahl: This is fine but here's an alternative syntax if you prefer it:
```python
from datetime import date
str(date.today())
```

scripts/scrape.py
```diff
@@ -29,6 +29,8 @@
 parser.add_argument('--neo4j-server', type=str, default='localhost',
                     help='The FQDN to the Neo4j server')
 parser.add_argument('--kerberos', action='store_true', help='Use Kerberos for authentication')
+parser.add_argument('--until', type=str,
```
mprahl: Optional: Putting this under the `--since` argument would be nice for readability.

yashvardhannanavati:
@mprahl, I have made the changes :)

mprahl:
Nice! :+1:

PR #120, opened 2018-06-04
Fix tests to expect ContainerKojiBuild as a part of KojiBuild and Bug Fixes

yashvardhannanavati:


purview/models/__init__.py
```diff
@@ -19,6 +19,8 @@
 
 # Overrides from https://github.com/neo4j-contrib/neomodel/pull/327
 # These should be removed once a new version is released
+
```
mprahl: This seems unnecessary.

yashvardhannanavati: It was throwing a linting error :(

purview/utils/general.py
```diff
@@ -66,14 +66,20 @@ def inflate_node(result):
...
     from purview.models import names_to_model
 
-    for label in result.labels:
-        if label in names_to_model:
-            node_model = names_to_model[label]
-            break
+    if 'ContainerKojiBuild' in result.labels:
+        result_label = 'ContainerKojiBuild'
+    elif len(result.labels) > 1:
+        raise RuntimeError('inflate_node encounted a node with the multiple labels: {0}. '
```
mprahl: "with the" => "with"

purview/models/__init__.py
mprahl: Oh alright. I think if you rebase develop on master and rebase your branch on develop this should be resolved.

mprahl:
After the rebases occur and the typo is fixed. :+1:

PR #123, opened 2018-06-05
Fix exception handling

yashvardhannanavati:


mprahl:
CI failure seems to stem from mirror issues

PR #125, opened 2018-06-06
Fix a bug to extract build_id from xml in freshmaker scraper

yashvardhannanavati:


scrapers/freshmaker.py
```diff
@@ -80,9 +80,9 @@ def query_api_and_update_neo4j(self):
                     # TODO: Change this if a task can trigger multiple builds
                     build_id = xml_root.find(".//*[name='koji_builds'].//string")
 
-                    if build_id:
+                    if isinstance(build_id, ET.Element):
```
mprahl: What are the alternative types that `build_id` could be?

yashvardhannanavati: It can be None, and I tried using `if build_id` but it doesn't work for some reason.

mprahl: That's odd. How about this syntax then?
```python
try:
    build_id = xml_root.find(".//*[name='koji_builds'].//string").text
except AttributeError:
    build_id = None
```

yashvardhannanavati: It doesn't throw an attribute error when the element is not present. It returns None

mprahl: It would when you try to access `.text` on a value that doesn't have like `None`.

yashvardhannanavati: Ah, cool. I see. Thanks

mprahl:
:+1:

PR #128, opened 2018-06-07
Use date in SQL queries instead of datetime

mprahl:
This seems to improve the success rates of queries to Teiid but I'm not sure why.

### DIFF
--- a/estuary/utils/general.py
+++ b/estuary/utils/general.py
@@ -42,6 +42,18 @@ def timestamp_to_datetime(timestamp):
     raise ValueError(error_msg)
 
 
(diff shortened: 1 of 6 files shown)


yashvardhannanavati:
Instead of converting `datetime` to `date`, can we directly manipulate the `timestamp_to_datetime` function?

mprahl:
@yashvardhannanavati I thought about that but I want to keep it as is because we will need `timestamp_to_datetime` in the future when we eventually query Neo4j by date and time based on an API query

yashvardhannanavati:
:+1:

PR #130, opened 2018-06-08
 Use multi-processing in the dist-git scraper

sarah256:
This should be a workaround for the memory leak and be a performance boost.

### DIFF
--- a/scrapers/distgit.py
+++ b/scrapers/distgit.py
@@ -2,10 +2,8 @@
 
 from __future__ import unicode_literals
 import re
-from multiprocessing import Pool
+from multiprocessing.dummy import Pool as ThreadPool
 from os import getenv
-import json
-import gc
 
 from builtins import bytes
 from bs4 import BeautifulSoup
@@ -51,7 +49,7 @@ class DistGitScraper(BaseScraper):
 
         :param list results: a list of dictionaries
         """
-        pool = Pool(processes=8)
+        pool = ThreadPool(8)
         counter = 0
         for result in results:
             if counter % 200 == 0:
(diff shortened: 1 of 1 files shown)


scrapers/distgit.py
```diff
@@ -3,13 +3,15 @@
 from __future__ import unicode_literals
 import re
 from multiprocessing import Pool
+from multiprocessing.dummy import Pool as ThreadPool
 from os import getenv
-import json
-import gc
+from itertools import islice
```
ralphbean: Fancy!

scrapers/distgit.py
```diff
@@ -42,16 +44,50 @@ def run(self, since=None, until=None):
             end_date = timestamp_to_date(until)
         results = self.get_distgit_data(start_date, end_date)
         log.info('Successfully fetched {0} results from Teiid'.format(len(results)))
-        self.update_neo4j(results)
+        # Upload the results to Neo4j
+        _update_neo4j_partial = partial(
+            self._update_neo4j, neomodel_config.DATABASE_URL, len(results))
+        pool = Pool(2)
```
ralphbean: So, we have a multi-process pool of size two handing out chunks of work to two thread pools of size 8 below...

That's complicated!  It deserves an inline comment in the code for the next person who reads the code and doesn't notice that `Pool` is not the same as `ThreadPool`.

hanzz:
+1

scrapers/distgit.py
```diff
@@ -42,16 +44,50 @@ def run(self, since=None, until=None):
...
+        :param tuple counter_and_results: a tuple where the first index is the current counter and
+        the second index is a list of dictionaries representing results from Teiid
         """
-        pool = Pool(processes=8)
+        previous_total = counter_and_results[0]
+        results = counter_and_results[1]
+        # Since _update_neo4j will be run in a separate process, we must configure the database
+        # URL every time the method is run.
+        neomodel_config.DATABASE_URL = neo4j_url
+        pool = ThreadPool(8)
```
mprahl: Since we are using multi-processing above, this pool should be reduced to 4. Having 16 threads banging against cgit may be too much.

sarah256:
Addressed feedback and updated PR

mprahl:
:+1:

PR #133, opened 2018-06-12
Order Teiid queries in descending order in the scrapers

sarah256:
FACTORY-2593

### DIFF
--- a/scrapers/errata.py
+++ b/scrapers/errata.py
@@ -148,7 +148,7 @@ class ErrataScraper(BaseScraper):
             LEFT JOIN Errata_public.users AS reporter_users
                 ON main.reporter_id = reporter_users.id
             WHERE main.updated_at >= '{0}' AND main.updated_at <= '{1}'
-            ORDER BY main.id;
+            ORDER BY main.updated_at DESC;
         """.format(since, until)
         log.info('Getting Errata advisories since {0} until {1}'.format(since, until))
         return self.teiid.query(sql)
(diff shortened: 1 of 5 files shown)


scrapers/errata.py
```diff
@@ -187,7 +186,7 @@ def get_associated_builds(self, advisory_id):
             LEFT JOIN Errata_public.brew_builds AS brew_builds
                 ON brew_builds.id = brew_mappings.brew_build_id
             LEFT JOIN Errata_public.packages AS packages
-                ON brew_builds.package_id = packages.id WHERE errata_id = {0};
+                ON brew_builds.package_id = packages.id WHERE errata_id = {0}
```
yashvardhannanavati: This change isn't necessary, but if you want to do it, could you do it for all the scrapers and all the queries? 

Or could you add the `;` to other places where it is missing?

scrapers/errata.py
```diff
@@ -203,7 +202,7 @@ def get_attached_bugs(self, advisory_id):
         sql = """\
             SELECT filed_bugs.bug_id as id_
             FROM Errata_public.filed_bugs as filed_bugs
-            WHERE filed_bugs.errata_id = {0};
+            WHERE filed_bugs.errata_id = {0}
```
yashvardhannanavati: Same as above

scrapers/freshmaker.py
```diff
@@ -107,7 +107,6 @@ def query_api_and_update_neo4j(self):
                             raise
                         build.add_label(ContainerKojiBuild.__label__)
                         build = ContainerKojiBuild.get_or_create(build_params)[0]
-
```
yashvardhannanavati: This change isn't required. Could you please undo this?

scrapers/errata.py
```diff
@@ -148,7 +148,7 @@ def get_advisories(self, since, until):
             LEFT JOIN Errata_public.users AS reporter_users
                 ON main.reporter_id = reporter_users.id
             WHERE main.updated_at >= '{0}' AND main.updated_at <= '{1}'
-            ORDER BY main.id;
+            ORDER BY main.update_date DESC;
```
yashvardhannanavati: Could you please change this to `ORDER BY main.updated_at DESC;`. It's just that there were some inconsistencies with the `update_date` attribute the last time we checked. Sorry I didn't catch this earlier

PR #135, opened 2018-06-13
Add the installation requirements in setup.py

mprahl:


setup.py
```diff
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0+
 from setuptools import setup, find_packages
 
+requirements = []
+with open('requirements.txt', 'r') as f:
```
yashvardhannanavati: Since you are adding the installation requirements here, maybe we should also change the README which currently asks the user to install requirements manually after running setup.py

yashvardhannanavati: We could also do that in a separate PR if you prefer so.

PR #139, opened 2018-06-14
Raise an exception for syntax errors in queries

yashvardhannanavati:


mprahl:
:+1:

mprahl:
@yashvardhannanavati I actually realized that you needed to add that if statement to the except around the query, not the connection.

PR #141, opened 2018-06-14
Move the syntax error checking part to query instead of connection

yashvardhannanavati:


yashvardhannanavati:
correcting my mistake, apologies!

mprahl:
:shipit:

PR #143, opened 2018-06-18
Fix bug to return the artifact itself when an artifact has no stories

yashvardhannanavati:


yashvardhannanavati:
So, we were previously returning an empty list of list if the query to Neo4j returns no result. This was causing the if condition on #189 to evaluate to `False` when checked which in turn returns an 'Internal Server Error' on the front-end. Whereas in this particular scenario that if condition should evaluate to `True`.  Hence, the fix :)

yashvardhannanavati:
@sarah256 A chance to understand the all_stories API endpoint if you want :)

mprahl:
:+1:

PR #145, opened 2018-06-19
Allow searching by more identifiers

mprahl:


estuary/models/koji.py
```diff
@@ -28,6 +31,31 @@ class KojiBuild(EstuaryStructuredNode):
...
+        uid = identifier
+        if re.match(r'^\d+$', uid):
+            # The identifier is an ID
+            return cls.nodes.get_or_none(id_=uid)
+        elif uid.endswith('.src.rpm'):
+            # The identifer is likely an NVR with .src.rpm at the end, so strip that part of it
+            # so it can be treated like a normal NVR
+            uid = uid[:-8]
+
+        if len(uid.rsplit('-', 2)) == 3:
```
yashvardhannanavati: Can we support this with another heuristic as well? This doesn't look bullet proof

tests/api/test_story.py
```diff
@@ -491,24 +491,24 @@
...
+                'security_impact': 'None',
+                'security_sla': None,
+                'state': 'SHIPPED_LIVE',
+                'states': [
 
                 ],
-                'status_time':'2017-08-01T15:43:51+00:00',
-                'synopsis':'cifs-utils bug fix update',
-                'triggered_freshmaker_event':[
+                'status_time': '2017-08-01T15:43:51+00:00',
```
yashvardhannanavati: @mprahl This is the reason I use sublime :P

estuary/models/koji.py
mprahl: I thought maybe there was, but I took the applicable code from:
https://github.com/release-engineering/kobo/blob/master/kobo/rpmlib.py#L166

yashvardhannanavati: cool, +1

PR #146, opened 2018-06-19
Add the siblings API endpoint and it's tests

yashvardhannanavati:


estuary/api/v1.py
```diff
@@ -218,3 +222,42 @@ def _get_partial_stories(query, reverse=False):
...
+    :param str uid: the value of the UniqueIdProperty to query with
+    :return: a Flask JSON response
+    :rtype: flask.Response
+    :raises NotFound: if the item is not found
+    :raises ValidationError: if an invalid resource was requested
+    """
+    # If last is True, we fetch siblings of the next node, previous node otherwise
+    last = request.args.get('last', False)
+
+    item = get_neo4j_node(resource, uid)
```
mprahl: I wonder if this is actually necessary query since you already know the label from `resource` and if no results are returned below, you know the item doesn't exist.

estuary/api/v1.py
```diff
@@ -218,3 +222,42 @@ def _get_partial_stories(query, reverse=False):
...
+    if not item:
+        raise NotFound('This item does not exist')
+
+    item_story_flow = story_flow(item.__label__)
```
mprahl: This could cause problems if the user requested a `KojiBuild` but `get_neo4j_node` returned a `ContainerKojiBuild`.

estuary/api/v1.py
```diff
@@ -218,3 +222,42 @@ def _get_partial_stories(query, reverse=False):
...
+
+    # Inflating and formatting results from Neo4j
+    serialized_results = []
+    for result in correlated_nodes:
+        inflated_node = inflate_node(result[0])
+        serialized_node = inflated_node.serialized
+        serialized_node['resource_type'] = inflated_node.__label__
+        serialized_results.append(serialized_node)
+
+    return jsonify(serialized_results) or []
```
mprahl: Remove ` or []`

estuary/api/v1.py
```diff
@@ -218,3 +222,42 @@ def _get_partial_stories(query, reverse=False):
...
+    if last:
+        correlated_nodes = get_correlated_node(item_story_flow['forward_label'], item, last=True)
+    else:
+        correlated_nodes = get_correlated_node(item_story_flow['backward_label'], item)
+
+    # Inflating and formatting results from Neo4j
+    serialized_results = []
+    for result in correlated_nodes:
+        inflated_node = inflate_node(result[0])
+        serialized_node = inflated_node.serialized
```
mprahl: We should do `serialized_all` so that way we can display things about relationships like the user associated with the artifact. We can reconsider if there is a performance hit.

estuary/api/v1.py
mprahl: Actually, you only use the label from the `item` so I think it could actually be refactored

estuary/api/v1.py
```diff
@@ -218,3 +222,42 @@ def _get_partial_stories(query, reverse=False):
...
+    item = get_neo4j_node(resource, uid)
+    if not item:
+        raise NotFound('This item does not exist')
+
+    item_story_flow = story_flow(item.__label__)
+    if not item_story_flow['backward_label'] or not item_story_flow['forward_label']:
+        raise ValidationError('Siblings cannot be determined on this kind of resource')
+
+    if last:
+        correlated_nodes = get_correlated_node(item_story_flow['forward_label'], item, last=True)
```
mprahl: In this context `get_correlated_node` doesn't sound like the proper name. Perhaps make this `get_correlated_nodes` and rename `get_correlated_nodes` to `get_all_correlated_nodes_count`.

estuary/api/v1.py
yashvardhannanavati: I do need the `item` because I am sending it to `get_correlated_nodes` and it uses the Neo4j id of that node. So I think we will have to keep this

estuary/api/v1.py
yashvardhannanavati: This would not cause problems because `get_neo4j_node` uses neomodel to query Neo4j and neomodel will only return the label you requested for. Hence, the problem you mention is taken care of inherently.

mprahl: Oh alright, perfect.

estuary/api/v1.py
mprahl: This is fine, it's just that in `get_correlated_nodes` you only use the label.

yashvardhannanavati: In lines #251 and #253, I send `item` as `next_node` to (now)`get_correlated_nodes`. And while querying Neo4j, we use the `id` in #238 of general.py

yashvardhannanavati:
@mprahl Rebased. :) It's ready for review again

yashvardhannanavati:
Oh, wait. `serialized_all` will change the expected outputs. I'll change the tests

estuary/api/v1.py
yashvardhannanavati: neomodel_flaws++ :p

estuary/utils/general.py
```diff
@@ -194,27 +194,30 @@ def get_correlated_nodes(results):
...
-    :param EstuaryStructuredNode curr_node: node for which the siblings count is to be calculated
+    :param str curr_node_label: node label for which the siblings count is to be calculated
     :param EstuaryStructuredNode next_node: correlated node to curr_node in the story/path
     :kwarg bool last: determines if it's the last node in the story/path
-    :return: siblings count of curr_node
-    :rtype: int
+    :kwarg bool last: determines if only count of sibling nodes should be returned
+    or the nodes themselves
+    :return: siblings count of curr_node | sibling nodes
+    :rtype: int | EstuaryStructuredNode
```
mprahl: It's actually returning the raw Neo4j results not `EstuaryStructuredNode`

PR #150, opened 2018-06-21
Fix an issue where a story ends with an avisory and the build is a container build

mprahl:


yashvardhannanavati:
As we discussed, this works for now but we have to find a permanent solution so that no more if statements are required when we add `ContainerAdvisory`. :+1:

PR #154, opened 2018-06-22
Set the multi-processing pool to 1 on the dist-git scraper to reduce RAM usage

mprahl:


scrapers/distgit.py
```diff
@@ -47,8 +47,9 @@ def run(self, since=None, until=None):
         # Upload the results to Neo4j
         _update_neo4j_partial = partial(
             self._update_neo4j, neomodel_config.DATABASE_URL, len(results))
-        # Create a multi-processing pool to process chunks of results
-        pool = Pool(2)
+        # Create a multi-processing pool to process chunks of results. This is set to 1 to reduce
+        # RAM usage.
+        pool = Pool(1)
```
yashvardhannanavati: Isn't Pool(1) equivalent to having a linear process? Or am I missing something

mprahl: Yes but the benefit of using multiprocessing is that a new a process is created for every 5000 entries and then the process is destroyed and the RAM is reallocated to the OS. This was an attempt to address the memory leak.

yashvardhannanavati: okay, sounds good :)

PR #158, opened 2018-06-25
Remove pusher from distgit scraper

sarah256:


scrapers/distgit.py
```diff
@@ -199,11 +184,9 @@ def get_distgit_data(self, since, until):
         """
         sql = """\
             SELECT c.commit_id, c.author, c.author_date, c.committer, c.commit_date, c.log_message,
-                c.sha, bz.bugzilla_id, bz.type as bugzilla_type, p.push_date, p.push_ip, p.module,
-                p.ref, p.push_id, p.pusher
+                c.sha, bz.bugzilla_id, bz.type as bugzilla_type, p.module,
+                p.ref
             FROM gitbz.git_commits as c
-            LEFT JOIN gitbz.git_push_commit_map as map ON c.commit_id = map.commit_id
-            LEFT JOIN gitbz.git_pushes as p ON p.push_id = map.push_id
```
yashvardhannanavati: If you remove this `LEFT JOIN`, you would not be able to retrieve p.module and p.ref. Hence, we would probably need this statement

yashvardhannanavati:
Also, could you please remove the `DistGitPush` model class from estuary/models/distgit.py

yashvardhannanavati:
So it looks like there are references of `DistGitPush` in other files. I would recommend doing a `git grep DistGitPush` in the root folder and then removing any imports that have `DistGitPush`. For more details, please check log of the failed build.

tests/api/test_get_resource.py
```diff
@@ -500,11 +500,6 @@ def test_get_resources(client, resource, uid, expected):
         'repo_name': 'some_repo_name',
         'repo_namespace': 'some_repo_namespace'
     })[0]
-    push = DistGitPush.get_or_create({
```
yashvardhannanavati: Looks like this file still has references to this push object in relationship definition and expected output above. Could you please remove them and try running the tests locally by first doing a `systemctl start docker` and then `sudo ./scripts/run-tests.sh` to see if they all pass?

yashvardhannanavati:
:shipit:

PR #159, opened 2018-06-25
Overwrite results with the formatted results in the dist-git scraper

mprahl:
The dist-git scraper stores the results from the Teiid query in the `results` variable. Then it used to run `pool.map` on a method that formatted the results and returned a generator object. The issue is that it seems `pool.map` exhausts the generator right away, causing some data to be duplicated unnecessarily (not sure how much in the results variable and the formatted results point to the same objects in RAM). This may be negligible, but I think it's worth doing to save what we can.

@sarah256 could you please review this?


sarah256:
:+1:

PR #160, opened 2018-06-25
Refactor the dist-git scraper to not reuse processes while multi-processing

mprahl:
It turns out that using a multiprocess.Pool will reuse the processes rather than
destroy it and create a new one. This makes it so that the memory leak in one of
the libraries is persisting. This commit refactors this so that processes are
terminated after each chunk and created for every chunk processed.

It turns out, the memory consumption is coming from large diffs in cgit. The last commit also addresses that by using a query parameter to stop showing the diff.

### DIFF
--- a/scrapers/distgit.py
+++ b/scrapers/distgit.py
@@ -101,90 +101,92 @@ class DistGitScraper(BaseScraper):
         :param tuple counter_and_results: a tuple where the first index is the current counter and
         the second index is a list of dictionaries representing results from Teiid
         """
-        previous_total = counter_and_results[0]
-        results = counter_and_results[1]
-        # Since _update_neo4j will be run in a separate process, we must configure the database
-        # URL every time the method is run.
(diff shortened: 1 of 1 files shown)


mprahl:
@sarah256 could you please review?

mprahl:
So I ran into this issue while running it for a while:
https://github.com/neo4j/neo4j-python-driver/issues/236

We'll need to figure out how to drop the connections after each thread.

yashvardhannanavati:
:+1:

sarah256:
:+1:

mprahl:
I'll keep debugging to fix the issue I've been encountering where one of the processes hangs and starts consuming a lot of RAM. It's very unpredictable when this happens.

mprahl:
I added a couple commits. @sarah256 could you please review?

sarah256:
:+1:

PR #161, opened 2018-06-27
Handle label ordering of stories

yashvardhannanavati:
So this hopefully gets rid of all the extra if statements we used to handle ordering of artifacts having multiple labels. For example, a build can have two labels, `KojiBuild` and  `ContainerKojiBuild`. Now, if the user requests for a kojibuild with id 123, and if that build has two labels, then our code can possibly show the build requested by the user as a `ContainerKojiBuild`. Hence, there can be inconsistencies in the output shown by estuary. This PR resolves this problem by comparing the results returned by our Neo4j graph database with a pre-defined story flow order and ensuring expected labels are assigned to their corresponding node.

I hope this helps. Happy reviewing! :)

### DIFF
--- a/estuary/utils/general.py
+++ b/estuary/utils/general.py
@@ -214,21 +214,7 @@ def get_correlated_nodes(curr_node_label, next_node, last=False, count=False):
     :return: siblings count of curr_node | sibling nodes
     :rtype: int | EstuaryStructuredNode
     """
-    # To avoid circular imports
-    from estuary.models.koji import KojiBuild
-    from estuary.models.distgit import DistGitCommit
-
     next_node_label = next_node.__label__
-    if curr_node_label == DistGitCommit.__label__:
-        # Always consider the next node as a KojiBuild because if it's
-        # a ContainerKojiBuild after a DistGitCommit, it should be
-        # treated as a normal KojiBuild in the story flow
-        next_node_label = KojiBuild.__label__
-    elif last and next_node_label == 'ContainerKojiBuild':
-        # Always consider the next node as a KojiBuild because if it's
-        # a ContainerKojiBuild and the story ended on an Advisory it should be
-        # treated as a normal KojiBuild in the story flow
-        next_node_label = KojiBuild.__label__
     item_story_flow = story_flow(next_node_label)
     relationship = item_story_flow['backward_relationship'][:-1]
     if last:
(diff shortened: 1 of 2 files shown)


estuary/utils/general.py
```diff
@@ -366,3 +352,32 @@ def format_story_results(results, requested_item):
...
+    :param list results: nodes in a story/path
+    :kwarg bool reverse: determines if the results are in reverse order of the story flow
+    :return: results with story/path labels
+    :rtype: list
+    """
+    # Avoid circular imports
+    from estuary.models import story_flow_list
+
+    if results:
+        idx = 0
```
ralphbean: Cosmetic.  This could become `idx = 0 if not reverse else (len(results) - 1)`.  Not sure if that is more or less readable.

estuary/utils/general.py
```diff
@@ -366,3 +352,32 @@ def format_story_results(results, requested_item):
...
+    :param string requested_node_label: label of the node requested by the user
+    :param list results: nodes in a story/path
+    :kwarg bool reverse: determines if the results are in reverse order of the story flow
+    :return: results with story/path labels
+    :rtype: list
+    """
+    # Avoid circular imports
+    from estuary.models import story_flow_list
+
+    if results:
```
ralphbean: Cosmetic.  This could become

```python
if not results:
    return results

idx = ...
```

... which would allow you to de-indent the entire function.

estuary/utils/general.py
```diff
@@ -366,3 +352,32 @@ def format_story_results(results, requested_item):
...
+        idx = 0
+        if reverse:
+            idx = len(results) - 1
+        node_idx = story_flow_list.index(requested_node_label)
+        while (0 <= idx < len(results)):
```
ralphbean: Another way to do this might be to say:

```python
# Increment or decrement through the list, depending.
delta = -1 if reverse else 1
while (0 <= idx < len(results)):
    results[idx].__label__ = story_flow_list[node_idx]
    idx += delta
    node_idx += delta
```

yashvardhannanavati: Yeah, I like this change

mprahl:
:+1: once TravisCI passes, feel free to merge.

yashvardhannanavati:
Thanks guys!

PR #167, opened 2018-07-03
Get correlated nodes count in the reverse direction as well

sarah256:
Links to https://projects.engineering.redhat.com/browse/FACTORY-2716

### DIFF
--- a/estuary/utils/general.py
+++ b/estuary/utils/general.py
@@ -182,43 +182,56 @@ def create_story_query(item, node_id, reverse=False, limit=False):
     return query
 
 
-def get_correlated_nodes_count(results):
+def get_correlated_nodes_count(results, reverse=False):
     """
     Iterate through the results and yield correlated nodes.
 
     :param list results: contains inflated results from Neo4j
+    :kwarg bool reverse: determines if we are traversing the story/path backwards
     :return: yield the results count (int) received from Neo4j
     :rtype: generator
     """
-    if len(results) < 2:
+    len_story = len(results)
+    if len_story < 2:
         raise RuntimeError('This function can\'t be called with one or zero elements')
-    for index in range(len(results)):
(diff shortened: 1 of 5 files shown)


estuary/utils/general.py
```diff
@@ -182,43 +182,53 @@ def create_story_query(item, node_id, reverse=False, limit=False):
...
+                results[index].__label__, results[index + 1], count=True))
+    else:
+        for index in range(len_story - 1, 0, -1):
+            correlated_nodes.append(get_correlated_nodes(
+                results[index].__label__, results[index - 1], reverse=True, count=True))
+
+    # When traversing the story, the last node is skipped because there is no next node for it, so
+    # we must add a value of 0 as a placeholder
+    correlated_nodes.append(0)
+    return correlated_nodes
```
mprahl: If `reverse` is `True`, please reverse the list so it is in the correct order rather than `format_story_results`. It'd make things more readable.

estuary/utils/general.py
```diff
@@ -182,43 +182,53 @@ def create_story_query(item, node_id, reverse=False, limit=False):
...
-        else:
-            yield get_correlated_nodes(results[index].__label__, results[index + 1], count=True)
 
+    correlated_nodes = []
+    if not reverse:
+        for index in range(len_story - 1):
+            correlated_nodes.append(get_correlated_nodes(
+                results[index].__label__, results[index + 1], count=True))
+    else:
+        for index in range(len_story - 1, 0, -1):
```
mprahl: Add a comment saying you are iterating over the results backwards for convenience but it'll be reversed to be in the correct order later on.

yashvardhannanavati:
Thank you guys!

yashvardhannanavati:
But wait, I think we won't be able to merge this until the corresponding changes are made to the front-end. Merging this will break the front-end since we altered the structure in which data is returned

PR #170, opened 2018-07-10
Add relationships API endpoint to get one-to-many rels of an artifact

yashvardhannanavati:


estuary/api/v1.py
```diff
@@ -274,3 +274,35 @@ def get_siblings(resource, uid):
...
+    :rtype: flask.Response
+    :raises NotFound: if the item is not found
+    :raises ValidationError: if an invalid resource was requested
+    """
+    item = get_neo4j_node(resource, uid)
+    if not item:
+        raise NotFound('This item does not exist')
+
+    # relationship to be expanded
+    rel_name = request.args.get('rel_name')
```
mprahl: Since this is a required argument, it may be best to just put it in the route like:
```
/relationships/<resource>/<uid>/<relationship>
```

estuary/api/v1.py
```diff
@@ -274,3 +274,35 @@ def get_siblings(resource, uid):
...
+    if not rel_name or (rel_name not in [rel[0] for rel in item.__all_relationships__]):
+        raise ValidationError(
+            'Please provide a valid relationship name for {0} with uid {1}'.format(resource, uid))
+
+    related_nodes = eval('item.{0}.match()'.format(rel_name))
```
mprahl: How about?
```python
related_nodes = getattr(item, rel_name).match()
```

yashvardhannanavati:
@mprahl I have made the changes and added unit tests. Could you please review this again?

PR #171, opened 2018-07-12
Fix incorrect commit creation

yashvardhannanavati:
Relates to FACTORY-2697

### DIFF
--- a/scrapers/koji.py
+++ b/scrapers/koji.py
@@ -71,9 +71,9 @@ class KojiScraper(BaseScraper):
             LEFT JOIN events ON build.create_event = events.id
             LEFT JOIN package ON build.pkg_id = package.id
             LEFT JOIN brew.users ON build.owner = brew.users.id
-            WHERE events.time IS NOT NULL AND events.time >= '{0}' AND events.time <= '{1}'
+            WHERE build.id = '716789'
             ORDER BY build.start_time DESC;
-            """.format(start_date, end_date)
+            """
 
         return self.teiid.query(sql=sql_query)
 
@@ -215,12 +215,14 @@ class KojiScraper(BaseScraper):
                 continue
             # Getting task related to the current build
             task_dict = self.get_task(task_id)[0]
-            xml_root = ET.fromstring(task_dict['request'])
             commit_hash = None
-            for child in xml_root.iter('string'):
-                if child.text and child.text.startswith('git'):
-                    commit_hash = child.text.rsplit('#', 1)[1]
-                    break
+            # Only look for the commit hash if the build is an RPM or container
+            if task_dict['method'] in ('build', 'buildContainer'):
+                xml_root = ET.fromstring(task_dict['request'])
+                for child in xml_root.iter('string'):
+                    if child.text and child.text.startswith('git'):
+                        commit_hash = child.text.rsplit('#', 1)[1]
+                        break
 
             if not task_dict:
                 # Continue if no corresponding task found


scrapers/koji.py
```diff
@@ -218,9 +218,12 @@ def update_neo4j(self, builds):
             xml_root = ET.fromstring(task_dict['request'])
             commit_hash = None
             for child in xml_root.iter('string'):
+                # Avoids incorrect creation of Commits
+                if '.tar' in child.text:
+                    commit_hash = None
+                    break
                 if child.text and child.text.startswith('git'):
                     commit_hash = child.text.rsplit('#', 1)[1]
-                    break
```
mprahl: You want to keep this break statement.

scrapers/koji.py
```diff
@@ -218,9 +218,12 @@ def update_neo4j(self, builds):
             xml_root = ET.fromstring(task_dict['request'])
             commit_hash = None
             for child in xml_root.iter('string'):
+                # Avoids incorrect creation of Commits
+                if '.tar' in child.text:
+                    commit_hash = None
+                    break
                 if child.text and child.text.startswith('git'):
```
mprahl: Change this to an else if for clarity.

scrapers/koji.py
```diff
@@ -218,9 +218,12 @@ def update_neo4j(self, builds):
             xml_root = ET.fromstring(task_dict['request'])
             commit_hash = None
             for child in xml_root.iter('string'):
+                # Avoids incorrect creation of Commits
+                if '.tar' in child.text:
```
mprahl: This is to vague of a search. What if some metadata is stuffed in there that is unrelated and has ".tar" in it?

yashvardhannanavati: So I checked a couple of other tasks and they only had the commit hash so I decided to go ahead with this. I even looked through the data from teiid and there doesn't seem to be anything else which can distinguish a tar build. I am open to other ideas :)

scrapers/koji.py
yashvardhannanavati: I removed the break statement because what if our other condition to check if there's a .tar is after we extract the hash (which isn't really hash)

mprahl: Oh I see... I think there's probably a better way to do this. Let me dig into it.

yashvardhannanavati: Another solution I can think of is after extracting the hash, check if it has special characters like '/' or '-'. Or if there's a specified length of the hash then we can check for that too. WDYT @mprahl ?

yashvardhannanavati:
@mprahl We would not be able to use the approach you mentioned because the even the task with erroneous commit hash has this text `git://pkgs.devel.redhat.com/rpms/rhosp-director-images?#origin/rhos-10.0-rhel-7`

mprahl:
@yashvardhannanavati but isn't that task not a "build" or "buildContainer" method? So my suggestion should work since it only looks at RPM builds and container builds.

yashvardhannanavati:
@mprahl what are your thoughts on checking the `method` attribute instead which is present for all the tasks.

mprahl:
@yashvardhannanavati, yes that is what I suggested in the diff a few comments above.

yashvardhannanavati:
In the diff, you're checking for the `methodName` in the xml. I am suggesting the `method` attribute which should be present in the task_dict

mprahl:
@yashvardhannanavati sure, either way is fine.

PR #172, opened 2018-07-12
Install Estuary with pip so that the no-deps option can be used

mprahl:
Install Estuary with pip so that the no-deps option can be used to not install deps from PyPi if they are missing.


mikebonnet:
+1 Looks reasonable.

PR #173, opened 2018-07-12
Use `setup.py develop --no-deps`

mprahl:
Use `setup.py develop --no-deps` instead of `pip3 install . --no-deps` to support dynamic imports used in the `neomodel_install_labels` script that runs after every deployment.

Unfortunately, `setup.py install` can't be used because `--no-deps` is not an option, so `setup.py develop` is used instead. There isn't really any harm in doing this, but it's not ideal.

Alternatively, I could revert to using `setup.py install` but PyPi deps would be pulled in if they were missed during the RPM installation.


mikebonnet:
+1 Agreed it's not the prettiest, but it sounds like it gets the job done. Thank you for the clear explanation!

PR #174, opened 2018-07-17
Alter the format of data returned by Siblings API endpoint to contain…

sarah256:
… label for Table view

@yashvardhannanavati could you please review :)

### DIFF
--- a/estuary/api/v1.py
+++ b/estuary/api/v1.py
@@ -9,10 +9,11 @@ from neomodel import db
 from estuary import version
 from estuary.models import story_flow_list
 from estuary.models.base import EstuaryStructuredNode
+from estuary.models.distgit import DistGitCommit
 
 from estuary.utils.general import (
     str_to_bool, get_neo4j_node, create_story_query, story_flow, format_story_results,
-    get_correlated_nodes, inflate_node, set_story_labels
+    get_correlated_nodes, inflate_node, set_story_labels, create_header
 )
 from estuary.error import ValidationError
 
@@ -256,7 +257,8 @@ def get_siblings(resource, uid):
     item = get_neo4j_node(resource, uid)
     if not item:
         raise NotFound('This item does not exist')
-
+    # Reverse means down --maybe dirn instead
+    # Not reverse means up
     item_story_flow = story_flow(item.__label__)
     # If reverse is True, we fetch siblings of the next node, previous node otherwise
     reverse = str_to_bool(request.args.get('reverse'))
@@ -276,7 +278,19 @@ def get_siblings(resource, uid):
         serialized_node['resource_type'] = inflated_node.__label__
         serialized_results.append(serialized_node)
 
-    return jsonify(serialized_results)
+    if item.__label__ == DistGitCommit.__label__:
+        header = create_header(item.__label__, item.hash_, item_story_flow, reverse)
+    else:
+        header = create_header(item.__label__, item.id_, item_story_flow, reverse)
+
+    result = {
+        'data': serialized_results,
+        'meta': {
+            'siblings_table_header': header
+        }
+    }
+
+    return jsonify(result)
 
 
 @api_v1.route('/relationships/<resource>/<uid>/<relationship>')
--- a/estuary/utils/general.py
+++ b/estuary/utils/general.py
@@ -395,3 +395,33 @@ def set_story_labels(requested_node_label, results, reverse=False):
         idx += delta
         node_idx += delta
     return results
+
+
+def create_header(item_type, item_id, item_story_flow, reverse):
+    """
+    Generate a header for the siblings chart.
+
+    :param string item_type: the name of the type of current item
+    :param string item_id: the id of the current item
+    :param dict item_story_flow: has forward and backward relationships of current item
+    :param bool reverse: determines if the results are in the reverse order of the story flow
+    :return: returns the appropriate siblings chart header
+    :rtype: string
+    """
+    def _plural_label(label):
+        if label.endswith('y'):
+            label = label[:-1] + 'ies'
+        else:
+            label = label + 's'
+        return label
+
+    if reverse:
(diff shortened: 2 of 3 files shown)


estuary/api/v1.py
```diff
@@ -276,7 +276,18 @@ def get_siblings(resource, uid):
...
 
-    return jsonify(serialized_results)
+    item_type = item.__label__
+    item_id = item.id_
+    header = create_header(item_type, item_id, item_story_flow, reverse)
+
+    result = {
+        'data': serialized_results,
+        'meta': {
+            'header': header
```
yashvardhannanavati: could you please change the key to something more specific like `siblings_table_header`

estuary/api/v1.py
```diff
@@ -276,7 +276,18 @@ def get_siblings(resource, uid):
         serialized_node['resource_type'] = inflated_node.__label__
         serialized_results.append(serialized_node)
 
-    return jsonify(serialized_results)
+    item_type = item.__label__
+    item_id = item.id_
```
yashvardhannanavati: Creation of new variables, `item_type` and `item_id` seem unnecessary to me. Instead, you can directly pass them

estuary/api/v1.py
```diff
@@ -276,7 +276,16 @@ def get_siblings(resource, uid):
         serialized_node['resource_type'] = inflated_node.__label__
         serialized_results.append(serialized_node)
 
-    return jsonify(serialized_results)
+    header = create_header(item.__label__, item.id_, item_story_flow, reverse)
```
yashvardhannanavati: I think this will fail when the item is a DistGitCommit since it will have a `hash` and not `id`.

estuary/api/v1.py
```diff
@@ -276,7 +276,19 @@ def get_siblings(resource, uid):
         serialized_node['resource_type'] = inflated_node.__label__
         serialized_results.append(serialized_node)
 
-    return jsonify(serialized_results)
+    if item.id_:
+        header = create_header(item.__label__, item.id_, item_story_flow, reverse)
+    else:
+        header = create_header(item.__label__, item.hash, item_story_flow, reverse)
```
yashvardhannanavati: `item.hash` => `item.hash_`

estuary/api/v1.py
```diff
@@ -276,7 +276,19 @@ def get_siblings(resource, uid):
         serialized_node['resource_type'] = inflated_node.__label__
         serialized_results.append(serialized_node)
 
-    return jsonify(serialized_results)
+    if item.id_:
```
yashvardhannanavati: Checking this will throw an AttributeError. I would suggest using something like

 if item.__label__ == DistGitCommit.__label__:
    /* use item.hash_
else:
   /* use item.id_

estuary/utils/general.py
```diff
@@ -395,3 +395,24 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :return: returns the appropriate siblings chart header
+    :rtype: string
+    """
+    if reverse:
+        result = '{0}s triggered by {1} {2}'.format(
+            item_story_flow['backward_label'], item_type, item_id)
+        return result
+    else:
+        result = '{0}s that triggered {1} {2}'.format(
+            item_story_flow['forward_label'], item_type, item_id)
```
yashvardhannanavati: This should probably be `result = '{0}s that triggered {1} {2}'.format(item_story_flow['backward_label'], item_type, item_id)`

estuary/utils/general.py
```diff
@@ -395,3 +395,24 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :param string item_type: the name of the type of current item
+    :param string item_id: the id of the current item
+    :param dict item_story_flow: has forward and backward relationships of current item
+    :param bool reverse: determines if the results are in the reverse order of the story flow
+    :return: returns the appropriate siblings chart header
+    :rtype: string
+    """
+    if reverse:
+        result = '{0}s triggered by {1} {2}'.format(
+            item_story_flow['backward_label'], item_type, item_id)
```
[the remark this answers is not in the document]
yashvardhannanavati: Also, It would be nice if we could replace the static `triggered by` text with the `item_story_flow['forward_relationship'][:-1] `. But then you will have to take care of eliminating the caps and underscores (if present). 

What do you think @mprahl ?

estuary/utils/general.py
yashvardhannanavati: Same as above

tests/api/test_siblings.py
```diff
@@ -14,291 +14,314 @@
...
+                ],
+                'content_types': ['docker'],
+                'created_at': '2017-04-03T14:47:23+00:00',
+                'id': '12327',
+                'issue_date': '2017-08-01T05:59:34+00:00',
+                'product_name': 'Red Hat Enterprise Linux',
+                'product_short_name': 'RHEL',
+                'release_date': None,
+                'reporter': None,
+                'resource_type': 'ContainerAdvisory',
```
yashvardhannanavati: The `resource_type` returned by the API is `ContainerAdvisory` but the header says that `FreshmakerEvents are returned.`

yashvardhannanavati:
Just a couple of comments, everything else looks good. good job!

estuary/utils/general.py
```diff
@@ -395,3 +395,33 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    Generate a header for the siblings chart.
+
+    :param string item_type: the name of the type of current item
+    :param string item_id: the id of the current item
+    :param dict item_story_flow: has forward and backward relationships of current item
+    :param bool reverse: determines if the results are in the reverse order of the story flow
+    :return: returns the appropriate siblings chart header
+    :rtype: string
+    """
+    def _plural_label(label):
```
yashvardhannanavati: This is good! nice thinking

PR #175, opened 2018-07-19
Change the 'reverse' flag to 'forward' for clarity

sarah256:


estuary/api/v1.py
```diff
@@ -92,18 +92,18 @@ def get_resource_story(resource, uid):
         raise NotFound('This item does not exist')
 
     forward_query = create_story_query(item, item.id, limit=True)
-    backward_query = create_story_query(item, item.id, reverse=True, limit=True)
+    backward_query = create_story_query(item, item.id, forward=True, limit=True)
 
-    def _get_partial_story(query, reverse=False):
+    def _get_partial_story(query, forward=False):
```
mprahl: Could you please revert this parameter change back to the way it was? This is function is about getting both halves of the story and not about the sibilings.

estuary/api/v1.py
```diff
@@ -92,18 +92,18 @@ def get_resource_story(resource, uid):
         raise NotFound('This item does not exist')
 
     forward_query = create_story_query(item, item.id, limit=True)
-    backward_query = create_story_query(item, item.id, reverse=True, limit=True)
+    backward_query = create_story_query(item, item.id, forward=True, limit=True)
```
yashvardhannanavati: I think we can keep this `reverse` since the query is actually being formed in the reverse direction of the story when the flag is true

estuary/api/v1.py
yashvardhannanavati: +1 on Matt's comment

estuary/api/v1.py
```diff
@@ -92,18 +92,18 @@ def get_resource_story(resource, uid):
...
 
         if not results:
             return []
 
         # Assuming that if Path is the first result, then that's all we want to process
         results = [list(results[0][0].nodes)]
-        # Reverse will be true when it is a backward query to preserve the story order
-        if reverse:
+        # forward will be true when it is a backward query to preserve the story order
+        if forward:
```
yashvardhannanavati: same as above

mprahl:
@sarah256 could you please redo this PR and limit the scope of the changes to the siblings API endpoint and the `get_correlated_nodes` functions? The other code you changed don't need changing.

estuary/api/v1.py
```diff
@@ -154,9 +154,9 @@ def get_resource_all_stories(resource, uid):
             break
 
     forward_query = create_story_query(item, item.id)
-    backward_query = create_story_query(item, item.id, reverse=True)
+    backward_query = create_story_query(item, item.id, forward=True)
```
yashvardhannanavati: we can keep this `reverse` too. Same goes for everything in this method

estuary/utils/general.py
```diff
@@ -127,13 +127,13 @@ def get_neo4j_node(resource_name, uid):
                 raise ValidationError(error)
 
 
-def create_story_query(item, node_id, reverse=False, limit=False):
+def create_story_query(item, node_id, forward=False, limit=False):
```
yashvardhannanavati: same as above

estuary/utils/general.py
```diff
@@ -363,19 +363,19 @@ def format_story_results(results, requested_item):
...
             'story_related_nodes_backward': list(
-                get_correlated_nodes_count(results, reverse=True)),
+                get_correlated_nodes_count(results, forward=True)),
             'requested_node_index': requested_node_index
         }
     }
 
 
-def set_story_labels(requested_node_label, results, reverse=False):
+def set_story_labels(requested_node_label, results, forward=False):
```
yashvardhannanavati: same as above

PR #178, opened 2018-07-19
Refactor the siblings API endpoint and underlying functions

mprahl:
Please review this by commit because they contain a description of the individual changes that are part of this refactor.


yashvardhannanavati:
I was just wondering if we can add an example comment as we discussed in the office today. We can do it in a separate PR too. Otherwise LGTM!

PR #179, opened 2018-07-20
Format the data returned by the Siblings API endpoint

sarah256:


estuary/api/v1.py
```diff
@@ -281,7 +282,21 @@ def get_siblings(resource, uid):
...
+        header = create_header(
+            story_node.__label__, story_node.hash_, story_node_story_flow, backward)
+    else:
+        header = create_header(
+            story_node.__label__, story_node.id_, story_node_story_flow, backward)
+
+    result = {
+        'data': serialized_results,
+        'meta': {
+            'siblings_table_header': header
```
mprahl: Could you rename this key to `description` or something along those lines? `siblings_table_header` is very specific and this could be useful for other things.

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
         idx += delta
         node_idx += delta
     return results
+
+
+def create_header(item_type, item_id, story_node_story_flow, backward):
```
mprahl: Could you please rename this to something more descriptive like `get_siblings_title`? `create_header` sounds too much like you're creating HTTP headers for an API query.

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    """
+    Generate a header for the siblings chart.
```
mprahl: How about `Generate a description of the siblings.`?

estuary/utils/general.py
mprahl: `item_type` and `item_id` are non-descriptive. Also, why not just pass the `story_node` object directly instead of passing in two of it's parameters?

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+
+    :param string item_type: the name of the type of current item
```
mprahl: This needs to be more descriptive

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :param string item_id: the id of the current item
```
mprahl: This needs to be more descriptive

estuary/utils/general.py
mprahl: Optional: I know it's less efficient but I'd prefer this function actually get the story_node_story_flow value instead of having it passed in. It'd make the function call simpler and cleaner and the function to get that value is quite light-weight.

mprahl: `backward` should default to `False` for convenience

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :param dict story_node_story_flow: has forward and backward relationships of current item
+    :param bool backward: determines if the results are in the reverse order of the story flow
```
mprahl: How about `determines if the results are in the reverse order of the story flow` => `determines the relationship direction the passed in node has with the siblings in the story`?

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :return: returns the appropriate siblings chart header
```
mprahl: "chart header" => "title"

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :rtype: string
+    """
+    def _plural_label(label):
+        if label.endswith('y'):
+            label = label[:-1] + 'ies'
+        else:
+            label = label + 's'
+        return label
+
+    if backward and story_node_story_flow['forward_label']:
```
mprahl: This logic is not right. If `backward` is `True`, then the `backward_label` should be used and not `forward_label`.

mprahl: The code in these if blocks duplicate a lot. The only difference from what I can tell is using `forward` or `backward` in how you index certain dictionaries. You should be able to assign that value to a variable and then use the same code for both scenarios afterwards.

For instance:
```
story_node_story_flow['{0}_label'.format(rel_direction)]
```

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+        fwd_label = _plural_label(story_node_story_flow['forward_label'])
+        fwd_relationship = \
+            story_node_story_flow['forward_relationship'][:-1].lower().replace('_', ' ')
+        result = '{0} {1} {2} {3}'.format(fwd_label, fwd_relationship, item_type, item_id)
+        return result
+    elif story_node_story_flow['backward_label']:
```
mprahl: You have to check the value of `backward` again to make sure it's the opposite of the last if statement since your last if statement uses "and" logic.

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+            label = label[:-1] + 'ies'
+        else:
+            label = label + 's'
+        return label
+
+    if backward and story_node_story_flow['forward_label']:
+        fwd_label = _plural_label(story_node_story_flow['forward_label'])
+        fwd_relationship = \
+            story_node_story_flow['forward_relationship'][:-1].lower().replace('_', ' ')
+        result = '{0} {1} {2} {3}'.format(fwd_label, fwd_relationship, item_type, item_id)
```
mprahl: I don't think this will always be proper grammar based on the relationship names. You'll have to write a function that makes the relationship names consistent grammatically. Perhaps you could add these values in the return value of `story_flow` like `forward_relationship_display`.

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    """
+    def _plural_label(label):
+        if label.endswith('y'):
+            label = label[:-1] + 'ies'
+        else:
+            label = label + 's'
+        return label
+
+    if backward and story_node_story_flow['forward_label']:
+        fwd_label = _plural_label(story_node_story_flow['forward_label'])
```
mprahl: This will still not be correct because labels like `BugzillaBug` will return `BugzillaBugs`. We actually want something like `Bugzilla bugs`. Perhaps you could add the display values in the return value of `story_flow` like `forward_label`. You'd still need to use your `_plural_label` function though.

estuary/utils/general.py
```diff
@@ -399,3 +399,37 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    Generate a header for the siblings chart.
+
+    :param string item_type: the name of the type of current item
+    :param string item_id: the id of the current item
+    :param dict story_node_story_flow: has forward and backward relationships of current item
+    :param bool backward: determines if the results are in the reverse order of the story flow
+    :return: returns the appropriate siblings chart header
+    :rtype: string
+    """
+    def _plural_label(label):
```
mprahl: How about `_get_plural_label` to have a consistent naming scheme with the rest of the project?

estuary/utils/general.py
```diff
@@ -399,3 +434,39 @@ def set_story_labels(requested_node_label, results, reverse=False):
         idx += delta
         node_idx += delta
     return results
+
+
+def get_siblings_description(item_id, story_node_story_flow, backward):
```
mprahl: For clarity, could you please rename `item_id` to `story_node_id`?

estuary/utils/general.py
```diff
@@ -399,3 +434,39 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    """
+    Generate a description of the siblings.
+
+    :param string item_id: the id of the current story node
+    :param dict story_node_story_flow: has forward and backward relationships of current item
+    :param bool backward: determines the relationship direction the passed in node has with the
+                          siblings in the story
```
mprahl: This should be unindented to match the rest of the codebase.

estuary/utils/general.py
```diff
@@ -277,68 +277,103 @@ def story_flow(label):
...
+            'display_label': 'DistGit commit',
             'forward_relationship': '{0}<'.format(
                 DistGitCommit.koji_builds.definition['relation_type']),
+            'forward_relationship_display': 'built from',
             'forward_label': KojiBuild.__label__,
+            'forward_label_display': 'Koji build',
             'backward_relationship': '{0}>'.format(
                 DistGitCommit.resolved_bugs.definition['relation_type']),
-            'backward_label': BugzillaBug.__label__
+            'backward_relationship_display': 'resolved by',
```
mprahl: The wording is wrong because this is from the perspective of the DistGitCommit, not the BugzillaBug.

estuary/utils/general.py
```diff
@@ -277,68 +277,103 @@ def story_flow(label):
...
+            'backward_label': None,
+            'backward_label_display': None
         }
     elif label == DistGitCommit.__label__:
         return {
             'uid_name': DistGitCommit.hash_.db_property or DistGitCommit.hash.name,
+            'display_label': 'DistGit commit',
             'forward_relationship': '{0}<'.format(
                 DistGitCommit.koji_builds.definition['relation_type']),
+            'forward_relationship_display': 'built from',
```
mprahl: The wording is wrong because this is from the perspective of the DistGitCommit, not the KojiBuild.

estuary/utils/general.py
```diff
@@ -277,68 +277,103 @@ def story_flow(label):
...
+            'backward_label': DistGitCommit.__label__,
+            'backward_label_display': 'DistGit commit'
         }
     elif label == Advisory.__label__:
         return {
             'uid_name': Advisory.id_.db_property or Advisory.id.name,
+            'display_label': 'Advisory',
             'forward_relationship': '{0}<'.format(
                 Advisory.triggered_freshmaker_event.definition['relation_type']),
+            'forward_relationship_display': 'triggered by',
```
mprahl: The wording is wrong because this is from the perspective of the Advisory, not the FreshmakerEvent

estuary/utils/general.py
```diff
@@ -277,68 +277,103 @@ def story_flow(label):
...
             'forward_label': FreshmakerEvent.__label__,
+            'forward_label_display': 'Freshmaker event',
             'backward_relationship': '{0}>'.format(
                 Advisory.attached_builds.definition['relation_type']),
-            'backward_label': KojiBuild.__label__
+            'backward_relationship_display': 'attached to',
```
mprahl: The wording is wrong because this is from the perspective of the Advisory, not the KojiBuild.

estuary/utils/general.py
```diff
@@ -277,68 +277,103 @@ def story_flow(label):
...
+            'backward_label': KojiBuild.__label__,
+            'backward_label_display': 'Koji build'
         }
     elif label == FreshmakerEvent.__label__:
         return {
             'uid_name': FreshmakerEvent.id_.db_property or FreshmakerEvent.id.name,
+            'display_label': 'Freshmaker event',
             'forward_relationship': '{0}>'.format(
                 FreshmakerEvent.triggered_container_builds.definition['relation_type']),
+            'forward_relationship_display': 'triggered by',
```
mprahl: This wording is wrong. It should be `triggered`.

estuary/utils/general.py
```diff
@@ -277,68 +277,103 @@ def story_flow(label):
...
             'uid_name': ContainerAdvisory.id_.db_property or ContainerAdvisory.id.name,
+            'display_label': 'Container advisory',
             'forward_relationship': None,
+            'forward_relationship_display': None,
             'forward_label': None,
+            'forward_label_display': None,
             'backward_relationship': '{0}>'.format(
                 ContainerAdvisory.attached_builds.definition['relation_type']),
-            'backward_label': ContainerKojiBuild.__label__
+            'backward_relationship_display': 'attached to',
```
mprahl: The wording is wrong here because it is from the perspective of the ContainerAdvisory, not the ContainerKojiBuild

estuary/utils/general.py
```diff
@@ -399,3 +434,39 @@ def set_story_labels(requested_node_label, results, reverse=False):
         idx += delta
         node_idx += delta
     return results
+
+
+def get_siblings_description(story_node_id, story_node_story_flow, backward):
+    """
+    Generate a description of the siblings.
+
+    :param string story_node_id: the id of the current story node
```
mprahl: "the id of the current story node" => "the id of the current story node that the siblings are related to"

estuary/utils/general.py
```diff
@@ -399,3 +434,39 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :param dict story_node_story_flow: has forward and backward relationships of current item
```
mprahl: "of current item" => "of the story node"

estuary/utils/general.py
```diff
@@ -399,3 +434,39 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+    :param bool backward: determines the relationship direction the passed in node has with the
```
mprahl: "passed in node" => "story node"

estuary/utils/general.py
```diff
@@ -399,3 +434,39 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+        else:
+            label = label + 's'
+        return label
+
+    if backward:
+        rel_direction = 'backward'
+    else:
+        rel_direction = 'forward'
+
+    label = story_node_story_flow['{0}_label_display'.format(rel_direction)]
```
mprahl: Rename this to `rel_label` because `label` sounds like it's the story node's label.

estuary/utils/general.py
```diff
@@ -399,3 +434,39 @@ def set_story_labels(requested_node_label, results, reverse=False):
...
+
+    if label:
+        label = _get_plural_label(label)
+        relationship = \
+            story_node_story_flow['{0}_relationship_display'.format(rel_direction)]
+        return '{0} {1} {2} {3}'.format(
```
mprahl: We don't necessarily just want to display the UID of the story node. Let's leave it as is, but in a separate PR, work with me so that this changes to use the preferred display name (e.g. a Bugzilla bug is represented has `RHBZ#12345`). I have code written for this in the front-end.

tests/api/test_siblings.py
```diff
@@ -14,291 +14,315 @@
...
+                'security_sla': None,
+                'state': 'SHIPPED_LIVE',
+                'status_time': '2017-08-01T15:43:51+00:00',
+                'synopsis': 'cifs-utils bug fix update',
+                'triggered_freshmaker_event': [],
+                'update_date': '2017-08-01T07:16:00+00:00'
+            }
+        ],
+        'meta': {
+            'description': 'Container advisories attached to Container Koji build 710'
```
mprahl: The relationship is from the wrong perspective.

estuary/utils/general.py
sarah256: how should the wording be here, I'm unsure @mprahl

estuary/utils/general.py
sarah256: how should the wording be here, I'm unsure @mprahl

tests/api/test_siblings.py
sarah256: see above comment^^

estuary/models/distgit.py
```diff
@@ -33,6 +33,7 @@ class DistGitBranch(EstuaryStructuredNode):
 class DistGitCommit(EstuaryStructuredNode):
     """Definition of a dist-git commit in Neo4j."""
 
+    __label_display__ = 'commit'
```
yashvardhannanavati: The label display in bugzilla bug, freshmaker event, koji build starts with a capital letter while commit, advisory start with lower case. Could you please make it uniform

estuary/utils/general.py
yashvardhannanavati: So for example if it's from a perspective of a ContainerKojiBuild, the wording is correct since 'Container Koji builds are attached to a Container Advisory' and not the other way around. So for ContainerAdvisory, you can have something like 'Container Advisory having attached Container Koji Builds'. I couldn't find the exact wordings for the second statement. But I hope this helps

estuary/models/distgit.py
yashvardhannanavati: @sarah256 and I had a chat on IRC and she explained me the reason behind it. So I think it's fine as it is

mprahl: @sarah256 and @yashvardhannanavati we could consider title-casing it and then for @sarah256's purposes she could just call `.lower()` on the string. I'll leave this up to you both to decide.

estuary/models/koji.py
```diff
@@ -60,6 +61,7 @@ def find_or_none(cls, identifier):
 class ContainerKojiBuild(KojiBuild):
     """A Neo4j definition of a build that represents a container build in Koji."""
 
+    __label_display__ = 'container Koji build'
```
mprahl: Could you make this `container build` instead?

estuary/utils/general.py
mprahl: Either what @yashvardhannanavati suggested or you could make it something generic like `contains`.

estuary/models/koji.py
```diff
@@ -15,6 +15,7 @@
 class KojiBuild(EstuaryStructuredNode):
     """Definition of a Koji build in Neo4j."""
 
+    __label_display__ = 'Koji build'
```
mprahl: I'm thinking we should make this just `build` instead of `Koji build` but I could be swayed either way. @yashvardhannanavati opinions on this?

estuary/api/v1.py
```diff
@@ -281,7 +281,18 @@ def get_siblings(resource, uid):
...
 
-    return jsonify(serialized_results)
+    uid_name = story_node.unique_id_property
+    description = get_siblings_description(
+        getattr(story_node, uid_name), story_node_story_flow, backward)
+
+    result = {
+        'data': serialized_results,
+        'meta': {
+            'description': description[0].upper() + description[1:]
```
mprahl: The code making the first letter uppercase should likely be part of `get_siblings_description`.

mprahl:
@sarah256 is this ready for review again?

mprahl:
LGTM

PR #180, opened 2018-07-25
Use intuitive names in the description returned by the siblings API endpoint

sarah256:


tests/api/test_siblings.py
```diff
@@ -320,7 +320,7 @@
             }
         ],
         'meta': {
-            'description': 'Container advisories attached to container build 710'
+            'description': 'Container advisories attached to slf4j_2-1.7.4-4.el7_4_as'
```
mprahl: This is not correct. It should be `Container advisories that contain slf4j_2-1.7.4-4.el7_4_as` or something like that.

sarah256:
@mprahl  should be all good to merge :)

tests/utils/test_general.py
```diff
@@ -57,23 +57,26 @@ def test_timestamp_to_datetime_invalid(input_dt):
...
-    ('123456', 'ContainerKojiBuild', False, 'Container advisories attached to container build'
-        ' 123456'),
-    ('123456', 'ContainerKojiBuild', True, 'Freshmaker events that triggered container build'
-        ' 123456'),
-    ('123456', 'ContainerAdvisory', True, 'Container builds attached to container advisory 123456')
+@pytest.mark.parametrize('display_name,label,backward,expected', [
+    ('RHBZ#12345', 'BugzillaBug', False, 'Commits that resolved RHBZ#12345'),
+    ('commit #1234567', 'DistGitCommit', False, 'Builds built by commit #1234567'),
+    ('commit #1234567', 'DistGitCommit', True, 'Bugzilla bugs resolved by commit #1234567'),
+    ('python-requests-1.10-3', 'KojiBuild', False, 'Advisories attached to python-requests-1.10-3'),
```
mprahl: This text isn't quite right. It should be something like `Advisories that contain python-requests-1.10-3`.

mprahl:
:+1:

PR #184, opened 2018-07-31
Fix a key error

mprahl:
Fix a key error that would happen if the siblings description was requested for an invalid relationship. I know the error is not as descriptive as it was meant to be but I felt that we should not alter the function parameters to accommodate this error message that should never happen (it's a safe-guard if we have a logic error).

### DIFF
--- a/estuary/utils/general.py
+++ b/estuary/utils/general.py
@@ -462,5 +462,5 @@ def get_siblings_description(story_node_display_name, story_node_story_flow, bac
         result = '{0} {1} {2}'.format(rel_label, relationship, story_node_display_name)
         return result[0].upper() + result[1:]
     else:
-        raise RuntimeError('A node with the label {0} does not have a {1} relationship'.format(
-            story_node_story_flow['display_label'], rel_direction))
+        raise RuntimeError('{0} does not have a {1} relationship'.format(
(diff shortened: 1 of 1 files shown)


mprahl:
@sarah256 could you please review as well?

estuary/utils/general.py
```diff
@@ -462,5 +462,5 @@ def _get_plural_label(label):
         result = '{0} {1} {2}'.format(rel_label, relationship, story_node_display_name)
         return result[0].upper() + result[1:]
     else:
-        raise RuntimeError('A node with the label {0} does not have a {1} relationship'.format(
-            story_node_story_flow['display_label'], rel_direction))
+        raise RuntimeError('A node of this type does not have a {0} relationship'.format(
```
yashvardhannanavati: hm, can we use `story_node_display_name` instead?

mprahl: I was thinking that but I thought it'd add to the confusion.

yashvardhannanavati: I just feel not having the specific name might just add to confusion.

mprahl:
@yashvardhannanavati fixed

yashvardhannanavati:
:+1:

PR #187, opened 2018-08-01
Remove some uniqueness and required constraints on the Freshmaker label

mprahl:
Neomodel uses requires attributes as a way to determine if the node exists but we should restrict that to just the FreshmakerEvent ID in the event the data is not filled in entirely from a scraper or estuary-updater handler.

This also removes the useless url parameter


sarah256:
:+1:

PR #188, opened 2018-08-04
Add ModuleKojiBuild model and modify koji scraper to scrape for ModuleKojiBuilds

yashvardhannanavati:


yashvardhannanavati:
failing tests can be ignored for now as they will get fixed when the changes to the API endpoints are done. :)

sarah256:
:+1: looks good! im not sure about the SQL queries as im not too familiar with the db but otherwise good!

yashvardhannanavati:
rebased!

estuary/models/koji.py
```diff
@@ -69,9 +70,22 @@ class ContainerKojiBuild(KojiBuild):
         '.freshmaker.FreshmakerEvent', 'TRIGGERED', cardinality=ZeroOrOne)
 
 
+class ModuleKojiBuild(KojiBuild):
+    """ A Neo4j definition of a build that represents a module build in Koji. """
+
+    components = RelationshipTo('KojiBuild', 'ATTACHED')
+    content_koji_tag = RelationshipFrom('KojiTag', 'CONTAINS')
```
mprahl: This is a one-to-one mapping so could you set the cardinality accordingly?

estuary/models/koji.py
```diff
@@ -69,9 +70,22 @@ class ContainerKojiBuild(KojiBuild):
...
+    module_stream = StringProperty()
+    module_version = StringProperty()
+
+
 class KojiTag(EstuaryStructuredNode):
     """Definition of a Koji tag in Neo4j."""
 
     builds = RelationshipTo('KojiBuild', 'CONTAINS')
     id_ = UniqueIdProperty(db_property='id')
+    module_builds = RelationshipTo('ModuleKojiBuild', 'CONTAINS')
```
mprahl: This is a one-to-one mapping so could you set the cardinality accordingly?

scrapers/koji.py
```diff
@@ -113,6 +113,70 @@ def get_build_tags(self, build_id):
...
+
+    def process_tag_builds(self, tag_builds_info):
+        """
+        Update Neo4j with builds attached to a tag.
+
+        :param dict tag_builds_info: build id's associated to a tag
+        """
+        builds = []
+        for item in tag_builds_info:
+            build = self.get_build_info(item['build_id'])
```
mprahl: You could change `get_build_info` to accept a list of build IDs and then you could consolidate this for loop to one SQL query with `WHERE build.id = 12345 OR build.id = 23456....`

scrapers/koji.py
```diff
@@ -113,6 +113,70 @@ def get_build_tags(self, build_id):
...
+            if not build:
+                continue
```
mprahl: Maybe add a `log.warn` here?

scrapers/koji.py
```diff
@@ -150,6 +219,25 @@ def update_neo4j(self, builds):
...
+                try:
+                    build_params['context'] = module_extra_info.get('context')
+                    build_params['mbs_id'] = module_extra_info.get('module_build_service_id')
+                    build_params['module_name'] = module_extra_info.get('name')
+                    build_params['module_stream'] = module_extra_info.get('stream')
+                    build_params['module_version'] = module_extra_info.get('version')
+                    build = ModuleKojiBuild.create_or_update(build_params)[0]
+                except neomodel.exceptions.ConstraintValidationFailed:
+                    # This must have errantly been created as a KojiBuild instead of a
+                    # ContainerKojiBuild, so let's fix that.
```
mprahl: `ContainerKojiBuild` => `ModuleKojiBuild`

scrapers/koji.py
```diff
@@ -160,6 +248,23 @@ def update_neo4j(self, builds):
...
+                    module_build_tag = KojiTag.create_or_update(dict(
+                        id_=tag_info[0]['id'],
+                        name=module_build_tag_name
+                    ))[0]
+                    module_build_tag.module_builds.connect(build)
+
+                    for item in tag_info:
+                        tag_build = KojiBuild.get_or_create(dict(
+                            id_=item['build_id']
+                            ))[0]
```
mprahl: unindent this please

scrapers/koji.py
```diff
@@ -160,6 +248,23 @@ def update_neo4j(self, builds):
...
+                if module_build_tag_name:
+                    tag_info = self.get_tag_info(module_build_tag_name)
+                    module_build_tag = KojiTag.create_or_update(dict(
+                        id_=tag_info[0]['id'],
+                        name=module_build_tag_name
+                    ))[0]
+                    module_build_tag.module_builds.connect(build)
+
+                    for item in tag_info:
+                        tag_build = KojiBuild.get_or_create(dict(
```
mprahl: Can you rename `tag_build` to `module_component` or `component`?

scrapers/koji.py
yashvardhannanavati: Good Idea!

scrapers/koji.py
yashvardhannanavati: This concern goes away since we are using a consolidated SQL query and we wouldn't know if a build is missing and isn't returned

yashvardhannanavati:
@mprahl Fixed!

scrapers/koji.py
```diff
@@ -113,6 +113,71 @@ def get_build_tags(self, build_id):
...
+                build.task_id,
+                build.version
+            FROM build
+            LEFT JOIN events ON build.create_event = events.id
+            LEFT JOIN package ON build.pkg_id = package.id
+            LEFT JOIN brew.users ON build.owner = brew.users.id
+        """
+
+        sql_query += 'WHERE '
+        for index in range(len(build_ids)):
```
mprahl: This could be simplified with:
```python
'WHERE build.id = ' + ' OR build.id = '.join([str(build_id) for build_id in build_ids])
```

mprahl: Also, can you raise an exception if build_ids is empty

scrapers/koji.py
```diff
@@ -113,6 +113,71 @@ def get_build_tags(self, build_id):
...
+            sql_query += 'build.id = {}'.format(build_ids[index])
+
+            if index != (len(build_ids) - 1):
+                sql_query += ' OR '
+
+        return self.teiid.query(sql=sql_query)
+
+    def process_tag_builds(self, tag_builds_info):
```
mprahl: Optional: My preference would be to not make this a separate method since it's so short and used only in one place

scrapers/koji.py
```diff
@@ -160,6 +249,23 @@ def update_neo4j(self, builds):
             user = User.get_or_create(dict(username=username))[0]
             build.conditional_connect(build.owner, user)
 
+            if build.__label__ == ModuleKojiBuild.__label__:
+                module_build_tag_name = module_extra_info.get('content_koji_tag')
+                if module_build_tag_name:
+                    module_component = self.get_tag_info(module_build_tag_name)
```
mprahl: Can you rename `module_component` to `module_components`?

scrapers/koji.py
```diff
@@ -113,6 +113,71 @@ def get_build_tags(self, build_id):
...
+    def get_tag_info(self, tag_name):
+        """
+        Query Teiid for tag_id of a tag and build_ids associated to it.
+
+        :param str tag_name: tag name
+        :return: a list of dictionaries
+        :rtype: list
+        """
+        sql_query = """
+            SELECT tags.id, tag_listing.build_id
```
mprahl: Could return `tags.id` as `tag_id` for readability?

scrapers/koji.py
```diff
@@ -160,6 +249,23 @@ def update_neo4j(self, builds):
...
+                if module_build_tag_name:
+                    module_component = self.get_tag_info(module_build_tag_name)
+                    module_build_tag = KojiTag.create_or_update(dict(
+                        id_=module_component[0]['id'],
+                        name=module_build_tag_name
+                    ))[0]
+                    module_build_tag.module_builds.connect(build)
+
+                    for item in module_component:
+                        tag_build = KojiBuild.get_or_create(dict(
```
mprahl: Can you rename this to `module_component`?

mprahl:
I left a few comments. Once you decide what to do with them, :+1: .

scrapers/koji.py
yashvardhannanavati: I will add a condition to this method which will return an empty list if `build_ids` is empty

yashvardhannanavati:
@mprahl Updated the PR with changes! thank you for the suggestions, the code definitely looks more readable now

estuary/models/koji.py
```diff
@@ -69,9 +70,22 @@ class ContainerKojiBuild(KojiBuild):
         '.freshmaker.FreshmakerEvent', 'TRIGGERED', cardinality=ZeroOrOne)
 
 
+class ModuleKojiBuild(KojiBuild):
+    """A Neo4j definition of a build that represents a module build in Koji."""
+
+    components = RelationshipTo('KojiBuild', 'ATTACHED')
```
mprahl: I think the relationship should actually be `CONTAINS`. Sorry I missed this.

scrapers/koji.py
```diff
@@ -160,6 +238,25 @@ def update_neo4j(self, builds):
             user = User.get_or_create(dict(username=username))[0]
             build.conditional_connect(build.owner, user)
 
+            if build.__label__ == ModuleKojiBuild.__label__:
```
mprahl: Why isn't this part of the `elif self.is_module_build(build_dict):` block?

yashvardhannanavati: Because I wanted to keep that block explicit for checking if the build is created erroneously, just like ContainerKojiBuild. Also, since the change was relating to KojiTag's I wanted to keep it in the same section where other KojiTag's are handled.

estuary/models/koji.py
yashvardhannanavati: makes sense! I'll do it

PR #189, opened 2018-08-06
Remove KojiTasks from models and scraper

sarah256:


sarah256:
@mprahl and @yashvardhannanavati could you please review? :)

scrapers/koji.py
```diff
@@ -224,45 +205,6 @@ def update_neo4j(self, builds):
                         commit_hash = child.text.rsplit('#', 1)[1]
                         break
 
-            if not task_dict:
```
yashvardhannanavati: I am having doubts that line #198 will throw an error if an empty list is returned from teiid. But maybe it will never return an empty list because we are extracting the task_id from the build and the task will always be present. Should we make that statement fail safe? What are your thoughts @mprahl and @sarah256 ?

mprahl: That's a good point. This if statement should be moved to right after `task_dict = self.get_task(task_id)[0]`.

scrapers/koji.py
```diff
@@ -215,6 +196,10 @@ def update_neo4j(self, builds):
                 continue
             # Getting task related to the current build
             task_dict = self.get_task(task_id)[0]
+
```
yashvardhannanavati: Line #198 will still throw an error if the list returned is empty. 
I would suggest something like 
`try:`
    `self.get_task(task_id)[0]`
`except IndexError:`
    `continue`

PR #190, opened 2018-08-09
Add another heuristic to check for container builds

yashvardhannanavati:
So I checked all container builds whose `name` ends with "-container"(21214 builds). Then I checked builds which did not have the `image` attribute in their `extra` among those and found only 6 such builds which were actually rpm builds(pki-core-container each one) example: https://brew.engineering.redhat.com/brew/buildinfo?buildID=741245. 

Then I found out builds not having the `extra` attribute itself and found 769 such builds. I did random checks on few of those builds and turns out they were all built by osbs. Hence, concluding that they are in-fact container builds (example: https://brew.engineering.redhat.com/brew/buildinfo?buildID=676727).

So I think this change should fix the bug of identifying the container builds correctly.

@mprahl Please let me know what you think

### DIFF
--- a/scrapers/base.py
+++ b/scrapers/base.py
@@ -62,14 +62,13 @@ class BaseScraper(object):
         except (ValueError, TypeError):
             extra_json = {}
 
-        # Checking a heuristic for determining if a build is a container build since, currently
+        # Checking heuristics for determining if a build is a container build, since currently
         # there is no definitive way to do it.
         if extra_json and (extra_json.get('container_koji_build_id') or
                            extra_json.get('container_koji_task_id')):
             return True
-        # Checking another heuristic for determining if a build is a container build since
-        # currently there is no definitive way to do it.
-        elif (package_name.endswith('-container') or package_name.endswith('-docker')):
+        elif extra_json.get('image') and\
+                (package_name.endswith('-container') or package_name.endswith('-docker')):
             return True
         else:
             return False


scrapers/base.py
```diff
@@ -62,14 +62,16 @@ def is_container_build(self, build_info):
...
-        # Checking a heuristic for determining if a build is a container build since, currently
+        # Checking heuristics for determining if a build is a container build since, currently
         # there is no definitive way to do it.
         if extra_json and (extra_json.get('container_koji_build_id') or
                            extra_json.get('container_koji_task_id')):
             return True
-        # Checking another heuristic for determining if a build is a container build since
-        # currently there is no definitive way to do it.
-        elif (package_name.endswith('-container') or package_name.endswith('-docker')):
+        elif extra_json.get("image"):
```
mprahl: After talking with Luiz, I think we can change this to:
```python
elif extra_json.get('image') and (package_name.endswith('-container') or package_name.endswith('-docker')):
    return True
```

Then we can remove the rest of the if statements.

yashvardhannanavati: I think we should keep the one which checks for `container_koji_build_id`. Just to be sure

scrapers/base.py
```diff
@@ -62,14 +62,12 @@ def is_container_build(self, build_info):
         except (ValueError, TypeError):
             extra_json = {}
 
-        # Checking a heuristic for determining if a build is a container build since, currently
+        # Checking heuristics for determining if a build is a container build since, currently
```
mprahl: The comma in the comment is actually in the wrong place. I wouldn't have said anything except you're modifying that line :stuck_out_tongue:

scrapers/base.py
```diff
@@ -62,14 +62,12 @@ def is_container_build(self, build_info):
...
         # there is no definitive way to do it.
         if extra_json and (extra_json.get('container_koji_build_id') or
                            extra_json.get('container_koji_task_id')):
             return True
-        # Checking another heuristic for determining if a build is a container build since
-        # currently there is no definitive way to do it.
-        elif (package_name.endswith('-container') or package_name.endswith('-docker')):
+        elif extra_json.get('image') and (package_name.endswith('-container') or package_name.endswith('-docker')):
```
mprahl: flake8!

yashvardhannanavati: oops! `ctrl+shift+R` didn't work it seems:p

scrapers/base.py
yashvardhannanavati: hahaha, i'll fix it

mprahl:
:+1:

mprahl:
@yashvardhannanavati could you also fix this in estuary-updater?

yashvardhannanavati:
Thanks guys! :)

PR #191, opened 2018-08-10
Support multiple story flows for an artifact

yashvardhannanavati:
This PR refers to FACTORY-2851. 

A major restructuring. I have moved few methods which are used in creating a story from `utils/general.py` to a class named `StoryUtils` . This will enable us to support multiple story flows without many headaches (hopefully). Speaking of which, I have added support for the story flow which contains ModuleKojiBuild.

I have purposely kept `module_flow=False` in `StoryUtils` to make this a non-breaking change. We can change it to `True` once we have a PR up for FACTORY-2852. This does support our current workflow of standard stories without Modules. Hope this explanation helps!

@mprahl and @sarah256  Sorry in advance for such a huge PR!


estuary/utils/story.py
```diff
@@ -0,0 +1,391 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from __future__ import unicode_literals
+
+from neomodel import db
+
+from estuary.error import ValidationError
+from estuary.models import story_flow_list, module_story_flow_list
```
mprahl: Are these used anywhere else? It might make sense to define them on the class.

estuary/utils/story.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+class StoryUtils(object):
```
mprahl: Just an idea, how about calling this something like `StoryManager`?

estuary/utils/story.py
```diff
@@ -0,0 +1,391 @@
...
+    """A class containing utility methods to create a story for an artifact."""
+
+    def __init__(self, module_flow=False):
+        """
+        Instantiate variables required to get story of a node.
+
+        :kwarg bool module_flow: specifies if the story flow with modules is to
+        be taken for this artifact
+        """
+        self.story_order_list = module_story_flow_list
```
mprahl: Instead of a parameter of `module_flow` being passed in, how about converting this specific class to a base class and then you can have sub-classes for each story type. So you'd have a `ModuleStoryManager` class and a `ContainerStoryManager`. Each sub-class would define the story flow variable. In this event, the `ModuleStoryManger` class would likely inherit from `ContainerStoryManager` and then call the parent's `story_flow` method and update the return value.

mprahl:
@yashvardhannanavati I think your strategy in the `story_flow` method is spot on. I just would like to discuss the format of the class as I described in the above comment. I think making it more like a plugin system would be better.

yashvardhannanavati:
@mprahl Based on our talks, I think I have made the changes we discussed. This is ready for review again!

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from __future__ import unicode_literals
+
+import abc
```
mprahl: There should be a new line between these since `abc` is part of the standard library and `neomodel` is a third-party package.

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+    """A class containing utility methods to create a story for an artifact."""
+
+    @staticmethod
+    def get_story_manager(item, item_id, config, limit=False):
+        """
+        Select which story flow to follow.
+
+        :param node item: a Neo4j node whose story is requested by the user
+        :param int node_id: the internal Neo4j ID of the node
+        :kwarg bool limit: specifies if LIMIT keyword should be added to the created cypher query
```
mprahl: Shift this below the `:param flask...` line.

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+        :param flask.config.Config config: flask config
+        :return: stories of a node in forward and backward direction
```
mprahl: This needs to be updated.

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+        :rtype: list
+        """
+        for class_name in config.get('STORY_MANAGER_SEQUENCE'):
```
mprahl: Could you replace `config.get('STORY_MANAGER_SEQUENCE')` with `config.get('STORY_MANAGER_SEQUENCE', ['ContainerStoryManager'])` so we can default to just using the `ContianerStoryManager` if the config isn't set.

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+            story_manager = eval(class_name)()
```
mprahl: I hate eval because it runs arbitrary code. How about something like:
```python
story_manager_cls = globals().get(class_name)
if not story_manager_cls:
     # raise exception here
story_manager = story_manager_cls()
```

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+            story_manager.forward_story = story_manager.get_story_nodes(item, item_id, limit=limit)
+            story_manager.backward_story = story_manager.get_story_nodes(
+                item, item_id, reverse=True, limit=limit)
+            BaseStoryManager.story_flow_list = story_manager.story_flow_list
+            story = story_manager.forward_story + story_manager.backward_story
```
mprahl: I may just be confused but shouldn't backward be added to forward and not the other way around.

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+from estuary.models.distgit import DistGitCommit
+from estuary.models.errata import Advisory, ContainerAdvisory
+from estuary.models.freshmaker import FreshmakerEvent
+
+
+class BaseStoryManager(object):
+    """A class containing utility methods to create a story for an artifact."""
+
+    @staticmethod
+    def get_story_manager(item, item_id, config, limit=False):
```
mprahl: I don't think you need the parameter `item_id` because `item.id` will get you this value.

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+
+    def is_valid(self, story):
+        """
+        Determine if the story path matches the returned story.
+
+        :param str label: Neo4j node label
+        :return: whether story is valid for this story path
+        :rtype: bool
+        """
+        for model in [node[0][0].nodes for node in story]:
```
mprahl: Can you assign this to a variable:
```python
nodes = [node[0][0].nodes for node in story]
for node in nodes:
    if node.__label__ == ModuleKojiBuild.__label__:
        return True
return False
```

estuary/utils/story.py
```diff
@@ -0,0 +1,468 @@
...
+            idx = len(results) - 1
+        delta = -1 if reverse else 1
+        node_idx = self.story_flow_list.index(requested_node_label)
+        while (0 <= idx < len(results)):
+            results[idx].__label__ = self.story_flow_list[node_idx]
+            idx += delta
+            node_idx += delta
+        return results
+
+    def get_siblings_description(self, story_node_display_name, story_node_story_flow, backward):
```
mprahl: This should be a static method since you are passing in all the information and not using any information from the class.

estuary/utils/story.py
yashvardhannanavati: It doesn't matter because we are just using it to check if the required node is present to follow a particular story flow

estuary/utils/story.py
yashvardhannanavati: I feel it's just creating a useless variable which is used only once.

yashvardhannanavati:
@mprahl Made the changes! :)

estuary/utils/story.py
```diff
@@ -0,0 +1,473 @@
...
+        :param node item: a Neo4j node whose story is requested by the user
+        :param flask.config.Config config: flask config
+        :kwarg bool limit: specifies if LIMIT keyword should be added to the created cypher query
+        :return: instance of one of the story manager classes
+        :rtype: ModuleStoryManager/ContainerStoryManager
+        """
+        for class_name in config.get('STORY_MANAGER_SEQUENCE', ['ContainerStoryManager']):
+            story_manager_cls = globals().get(class_name)
+            if not story_manager_cls:
+                raise TypeError('')
```
mprahl: Could you add a message? Could you also make this a `RuntimeError`?

estuary/utils/story.py
```diff
@@ -0,0 +1,473 @@
...
+            story_manager = story_manager_cls()
+            story_manager.forward_story = story_manager.get_story_nodes(item, limit=limit)
+            story_manager.backward_story = story_manager.get_story_nodes(
+                item, reverse=True, limit=limit)
+            BaseStoryManager.story_flow_list = story_manager.story_flow_list
+            story = story_manager.forward_story + story_manager.backward_story
```
mprahl: I know you said it doesn't matter which order, but I'd prefer if the order was correct in the event `is_valid` is expanded to do more complex checks. That way we can avoid confusion in the future.

estuary/utils/story.py
```diff
@@ -0,0 +1,473 @@
...
+        """
+        for class_name in config.get('STORY_MANAGER_SEQUENCE', ['ContainerStoryManager']):
+            story_manager_cls = globals().get(class_name)
+            if not story_manager_cls:
+                raise TypeError('')
+            story_manager = story_manager_cls()
+            story_manager.forward_story = story_manager.get_story_nodes(item, limit=limit)
+            story_manager.backward_story = story_manager.get_story_nodes(
+                item, reverse=True, limit=limit)
+            BaseStoryManager.story_flow_list = story_manager.story_flow_list
```
mprahl: This is not used anymore as far as I can tell so this can be removed.

mprahl:
@yashvardhannanavati a few more comments. Let me know when they are addressed.

estuary/utils/story.py
```diff
@@ -0,0 +1,472 @@
...
+        Select which story flow to follow.
+
+        :param node item: a Neo4j node whose story is requested by the user
+        :param flask.config.Config config: flask config
+        :kwarg bool limit: specifies if LIMIT keyword should be added to the created cypher query
+        :return: instance of one of the story manager classes
+        :rtype: ModuleStoryManager/ContainerStoryManager
+        """
+        for class_name in config.get('STORY_MANAGER_SEQUENCE', ['ContainerStoryManager']):
+            story_manager_cls = globals().get(class_name)
```
mprahl: I'm sorry for being inconsistent here but after thinking about it, I think having a static variable that is a list of story manager classes would be better than just using `globals()`. Then in the error message below, you can say what the available classes are. This would also help in PR #192 with validating input from the user.

mprahl:
:+1:

PR #192, opened 2018-08-16
Support multiple story flows in siblings API endpoint

yashvardhannanavati:
This should only be merged after #191 is merged. 

PS: This will need a rebase after 191 is merged since this one builds on it.

### DIFF
--- a/estuary/api/v1.py
+++ b/estuary/api/v1.py
@@ -10,7 +10,7 @@ from estuary.models.base import EstuaryStructuredNode
 
 from estuary.utils.general import str_to_bool, get_neo4j_node, inflate_node
 from estuary.error import ValidationError
-from estuary.utils.story import BaseStoryManager, ContainerStoryManager
+import estuary.utils.story
 
 
 api_v1 = Blueprint('api_v1', __name__)
@@ -76,7 +76,7 @@ def get_resource_story(resource, uid):
     if not item:
         raise NotFound('This item does not exist')
 
-    story_manager = BaseStoryManager.get_story_manager(
+    story_manager = estuary.utils.story.BaseStoryManager.get_story_manager(
         item, current_app.config, limit=True)
 
     def _get_partial_story(results, reverse=False):
@@ -113,6 +113,7 @@ def get_resource_story(resource, uid):
         rv['meta']['story_related_nodes_forward'] = [0]
         rv['meta']['story_related_nodes_backward'] = [0]
         rv['meta']['requested_node_index'] = 0
+        rv['meta']['story_type'] = story_manager.__class__.__name__[:-12].lower()
         rv['data'][0]['resource_type'] = item.__label__
         rv['data'][0]['display_name'] = item.display_name
         return jsonify(rv)
@@ -140,7 +141,7 @@ def get_resource_all_stories(resource, uid):
         if item:
             break
 
-    story_manager = BaseStoryManager.get_story_manager(item, current_app.config)
+    story_manager = estuary.utils.story.BaseStoryManager.get_story_manager(item, current_app.config)
 
     def _get_partial_stories(results, reverse=False):
 
@@ -221,6 +222,7 @@ def get_resource_all_stories(resource, uid):
         rv['meta']['story_related_nodes_forward'] = [0]
         rv['meta']['story_related_nodes_backward'] = [0]
         rv['meta']['requested_node_index'] = 0
+        rv['meta']['story_type'] = story_manager.__class__.__name__[:-12].lower()
         rv['data'][0]['resource_type'] = item.__label__
         rv['data'][0]['display_name'] = item.display_name
         all_results.append(rv)
@@ -240,14 +242,22 @@ def get_siblings(resource, uid):
     :raises NotFound: if the item is not found
     :raises ValidationError: if an invalid resource was requested
     """
+    story_type_mapper = {'container': 'ContainerStoryManager', 'module': 'ModuleStoryManager'}
+    story_type = request.args.get('story_type', 'container').lower()
+
+    story_manager_class = story_type_mapper.get(story_type)
+    if story_manager_class:
+        story_manager = getattr(estuary.utils.story, story_manager_class)()
+    else:
+        raise ValidationError('Supplied story type is invalid. Select from: {0}'
+                              .format(', '.join(current_app.config['STORY_MANAGER_SEQUENCE'])))
+
     # This is the node that is part of a story which has the relationship to the desired
     # sibling nodes
     story_node = get_neo4j_node(resource, uid)
     if not story_node:
         raise NotFound('This item does not exist')
 
-    # Defaulting to Container Story until we modify this endpoint to suit the module changes
-    story_manager = ContainerStoryManager()
     story_node_story_flow = story_manager.story_flow(story_node.__label__)
     # If backward_rel is true, we fetch siblings of the previous node, next node otherwise.
     # For example, if you pass in an advisory and want to know the Freshmaker events triggered from
--- a/tests/api/test_all_stories.py
+++ b/tests/api/test_all_stories.py
@@ -101,7 +101,8 @@ from estuary.models.freshmaker import FreshmakerEvent
             'meta':{
                 'story_related_nodes_backward': [0, 0, 0, 1],
                 'story_related_nodes_forward': [0, 1, 0, 0],
-                'requested_node_index': 0
+                'requested_node_index': 0,
+                'story_type': 'container'
             }
         },
         {
@@ -215,7 +216,8 @@ from estuary.models.freshmaker import FreshmakerEvent
             'meta': {
                 'story_related_nodes_backward': [0, 0, 0, 1, 0, 1],
                 'story_related_nodes_forward': [0, 1, 0, 0, 0, 0],
-                'requested_node_index': 0
+                'requested_node_index': 0,
+                'story_type': 'container'
             }
         }
     ]),
@@ -339,7 +341,8 @@ from estuary.models.freshmaker import FreshmakerEvent
             'meta':{
                 'requested_node_index': 2,
                 'story_related_nodes_backward': [0, 0, 0, 1],
-                'story_related_nodes_forward':[0, 1, 0, 0]
+                'story_related_nodes_forward': [0, 1, 0, 0],
+                'story_type': 'container'
             }
         },
         {
@@ -461,7 +464,8 @@ from estuary.models.freshmaker import FreshmakerEvent
             'meta':{
                 'requested_node_index': 2,
                 'story_related_nodes_backward': [0, 0, 0, 1],
-                'story_related_nodes_forward':[0, 1, 0, 0]
+                'story_related_nodes_forward': [0, 1, 0, 0],
+                'story_type': 'container'
             }
         },
         {
@@ -608,7 +612,8 @@ from estuary.models.freshmaker import FreshmakerEvent
             'meta': {
                 'requested_node_index': 2,
                 'story_related_nodes_backward': [0, 0, 0, 1, 0, 1],
-                'story_related_nodes_forward':[0, 1, 0, 0, 0, 0]
+                'story_related_nodes_forward': [0, 1, 0, 0, 0, 0],
+                'story_type': 'container'
             }
         },
         {
@@ -755,7 +760,8 @@ from estuary.models.freshmaker import FreshmakerEvent
             'meta': {
                 'requested_node_index': 2,
                 'story_related_nodes_backward': [0, 0, 0, 1, 0, 1],
-                'story_related_nodes_forward':[0, 1, 0, 0, 0, 0]
+                'story_related_nodes_forward':[0, 1, 0, 0, 0, 0],
+                'story_type': 'container'
             }
         }
     ])
@@ -984,7 +990,8 @@ def test_get_stories_fallback(client):
         'meta': {
             'requested_node_index': 0,
             'story_related_nodes_forward': [0, 0],
-            'story_related_nodes_backward': [0, 0]
+            'story_related_nodes_backward': [0, 0],
+            'story_type': 'container'
         }
     }]
 
--- a/tests/api/test_siblings.py
+++ b/tests/api/test_siblings.py
@@ -6,7 +6,7 @@ from datetime import datetime
 
 import pytest
 
-from estuary.models.koji import KojiBuild, ContainerKojiBuild
+from estuary.models.koji import KojiBuild, ContainerKojiBuild, ModuleKojiBuild
 from estuary.models.bugzilla import BugzillaBug
 from estuary.models.distgit import DistGitCommit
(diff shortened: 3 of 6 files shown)


yashvardhannanavati:
Refers to FACTORY-2903

estuary/api/v1.py
```diff
@@ -10,7 +10,7 @@
 
 from estuary.utils.general import str_to_bool, get_neo4j_node, inflate_node
 from estuary.error import ValidationError
-from estuary.utils.story import BaseStoryManager, ContainerStoryManager
+from estuary.utils.story import BaseStoryManager, ContainerStoryManager, ModuleStoryManager  # NOQA
```
mprahl: Why `  # NOQA`?

estuary/api/v1.py
```diff
@@ -113,6 +113,7 @@ def _get_partial_story(results, reverse=False):
         rv['meta']['story_related_nodes_forward'] = [0]
         rv['meta']['story_related_nodes_backward'] = [0]
         rv['meta']['requested_node_index'] = 0
+        rv['meta']['story_manager'] = type(story_manager).__name__
```
mprahl: I'm torn about this because it is very convenient to just pass the class but at the same time, this is very much an implementation detail that shouldn't really be exposed in the API.

estuary/api/v1.py
```diff
@@ -240,14 +242,14 @@ def get_siblings(resource, uid):
...
+    story_manager_arg = request.args.get('story', 'ContainerStoryManager')
     # This is the node that is part of a story which has the relationship to the desired
     # sibling nodes
     story_node = get_neo4j_node(resource, uid)
     if not story_node:
         raise NotFound('This item does not exist')
-
     # Defaulting to Container Story until we modify this endpoint to suit the module changes
-    story_manager = ContainerStoryManager()
+    story_manager = globals().get(story_manager_arg)()
```
mprahl: This is quite dangerous because this allows the user to pass in a string of any function or class name and cause the back-end to execute it. My suggestion is that the argument should be something like `story_type` and then the back-end tries to find the matching story manager class. An example of input would be `container` or `module`.

estuary/api/v1.py
yashvardhannanavati: Because we use those classes when a parameter is passed in the siblings API. But they are not literally mentioned in the code anywhere and it gives an 'imported bu unused' error

estuary/api/v1.py
yashvardhannanavati: yeah, that makes sense. I will change it.

mprahl:
Looks good to me but we need some unit tests around this. Ping me when those are done and then I'll review and merge it.

estuary/api/v1.py
```diff
@@ -240,14 +242,21 @@ def get_siblings(resource, uid):
...
     # sibling nodes
     story_node = get_neo4j_node(resource, uid)
     if not story_node:
         raise NotFound('This item does not exist')
 
-    # Defaulting to Container Story until we modify this endpoint to suit the module changes
-    story_manager = ContainerStoryManager()
+    for class_name in current_app.config['STORY_MANAGER_SEQUENCE']:
+        if class_name.lower().startswith(story_type):
+            story_manager = getattr(estuary.utils.story, class_name)()
```
mprahl: This doesn't take into account if the config has an invalid class, but let's deal with that in a separate PR.

estuary/api/v1.py
```diff
@@ -240,14 +242,21 @@ def get_siblings(resource, uid):
...
     # This is the node that is part of a story which has the relationship to the desired
     # sibling nodes
     story_node = get_neo4j_node(resource, uid)
     if not story_node:
         raise NotFound('This item does not exist')
 
-    # Defaulting to Container Story until we modify this endpoint to suit the module changes
-    story_manager = ContainerStoryManager()
+    for class_name in current_app.config['STORY_MANAGER_SEQUENCE']:
+        if class_name.lower().startswith(story_type):
```
mprahl: Optional: I'm realizing now that this logic is fine for now but if we added another story path like `ModuleContainerStoryManager`, this would cause problems.

tests/api/test_siblings.py
```diff
@@ -40,18 +40,95 @@
...
+
+                ],
                 'name': 'slf3j',
                 'owner': None,
                 'release': '4.el6_3',
                 'resource_type': 'KojiBuild',
-                'display_name': 'slf3j-1.7.1-4.el6_3',
                 'start_time': '2017-04-02T19:39:06+00:00',
-                'state': 2,
+                'state':2,
```
mprahl: Can you add a space after the colon?

tests/api/test_siblings.py
```diff
@@ -368,6 +545,21 @@ def test_node_siblings(client, resource, uid, backward_rel, expected):
         'state': 2,
         'version': '1.7.1'
     })[0]
+    module_build = ModuleKojiBuild.get_or_create({
```
mprahl: How about only creating the module build and its relationships if the `story_type` is `module`. When that is the case, also avoid creating the relationships between the RPM builds and the advisory.

tests/api/test_siblings.py
```diff
@@ -6,15 +6,15 @@
...
+from estuary.models.koji import KojiBuild, ContainerKojiBuild, ModuleKojiBuild
 from estuary.models.bugzilla import BugzillaBug
 from estuary.models.distgit import DistGitCommit
 from estuary.models.errata import Advisory, ContainerAdvisory
 from estuary.models.freshmaker import FreshmakerEvent
 
 
-@pytest.mark.parametrize('resource,uid,backward_rel,expected', [
-    ('advisory', '27825', True, {
+@pytest.mark.parametrize('resource,uid,backward_rel,story_path,expected', [
```
mprahl: Instead of adding `story_path` as a parametrize parameter here, you could maybe add a second parametrize decorator that just deals with `story_path`. That way all existing scenarios are also tested with the `story_path` as module. If this is too difficult to read, you could break out the module `story_path` testing to a separate test function.

mprahl:
@yashvardhannanavati looks good but I left a couple of comments about the tests.

tests/api/test_siblings.py
```diff
@@ -501,3 +698,31 @@ def test_reverse_flag(client):
...
+        'product_short_name': 'RHEL',
+        'security_impact': 'None',
+        'state': 'SHIPPED_LIVE',
+        'status_time': datetime(2017, 8, 1, 15, 43, 51),
+        'synopsis': 'cifs-utils bug fix update',
+        'update_date': datetime(2017, 8, 1, 7, 16)
+    })[0]
+
+    expected = {
+        'message': 'Supplied story type is invalid. Select from: ModuleStoryManager, '
```
mprahl: Optional: I'd prefer something like:
```python
'message': ('Supplied story type is invalid. Select from: ModuleStoryManager, '
            'ContainerStoryManager'),
```

tests/api/test_story.py
```diff
@@ -1173,10 +1180,296 @@ def test_get_stories(client, resource, uids, expected):
     for uid in uids:
         url = '/api/v1/story/{0}/{1}'.format(resource, uid)
         rv = client.get(url)
-        assert rv.status_code == 200, 'Failed getting the resource at: {0}'.format(url)
+        assert rv.status_code == 200
         assert json.loads(rv.data.decode('utf-8')) == expected
 
 
+@pytest.mark.parametrize('resource,uids,expected', [
+    ('freshmakerevent', ['1180'], {
```
mprahl: You only use parametrize for one test, so this could be converted to a traditional test for better readability.

estuary/api/v1.py
yashvardhannanavati: I thought we agreed that since it's coming from config we don't need to deal with that

mprahl: Well we don't know if the classes in the config are valid. It's not that big of a deal because this isn't deployed by anyone but us, but it's still good practice.

tests/api/test_story.py
```diff
@@ -1173,10 +1180,296 @@ def test_get_stories(client, resource, uids, expected):
...
+        'start_time': datetime(2017, 4, 2, 19, 39, 6),
+        'state': 1,
+        'version': '1.7.4'
+    })[0]
+    module_build = ModuleKojiBuild.get_or_create({
+        'completion_time': datetime(2017, 4, 2, 19, 39, 6),
+        'creation_time': datetime(2017, 4, 2, 19, 39, 6),
+        'epoch': '0',
+        'id_': '2345',
+        'name': 'slf4j',
```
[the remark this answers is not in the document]
yashvardhannanavati: Those parameters which are common with a KojiBuild are taken from already existing builds in the test while those corresponding to a ModuleKojiBuild are actually taken from a brew build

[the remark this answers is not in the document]
mprahl: Exactly. The properties like name, version, and release are incorrect. They are not in the format of a module build.

tests/api/test_story.py
mprahl: Nevermind, we can keep this as is. It'll make it easier to expand later.

estuary/api/v1.py
yashvardhannanavati: In such case, I think we can then have a dict of valid parameters and their corresponding classes. Also the parameter in such a case would have to be something like 'module-container' or something similar. I think we can handle that using this logic too.

mprahl: @yashvardhannanavati a dictionary would be a good option. You can leave it as is or implement that.

estuary/api/v1.py
```diff
@@ -240,14 +242,21 @@ def get_siblings(resource, uid):
...
+    story_type = request.args.get('story_type', 'container').lower()
     # This is the node that is part of a story which has the relationship to the desired
     # sibling nodes
     story_node = get_neo4j_node(resource, uid)
     if not story_node:
         raise NotFound('This item does not exist')
 
-    # Defaulting to Container Story until we modify this endpoint to suit the module changes
-    story_manager = ContainerStoryManager()
+    story_manager_class = story_type_mapper.get(story_type)
```
mprahl: Can you move this code to before the API queries Neo4j (up 5 or so lines)?

tests/api/test_all_stories.py
```diff
@@ -339,7 +341,8 @@
             'meta':{
                 'requested_node_index': 2,
                 'story_related_nodes_backward': [0, 0, 0, 1],
-                'story_related_nodes_forward':[0, 1, 0, 0]
+                'story_related_nodes_forward':[0, 1, 0, 0],
```
mprahl: Where isn't there a space after the colon here?

tests/api/test_siblings.py
```diff
@@ -41,15 +41,17 @@
                 'commit': None,
                 'completion_time': '2017-04-02T19:39:06+00:00',
                 'creation_time': '2017-04-02T19:39:06+00:00',
+                'display_name': 'slf3j-1.7.1-4.el6_3',
                 'epoch': '0',
                 'extra': None,
                 'id': '3456',
-                'module_builds': [],
+                'module_builds':[
```
mprahl: Where isn't there a space after the colon here?

tests/api/test_siblings.py
```diff
@@ -87,24 +89,26 @@
...
+                ],
                 'name': 'slf4j',
                 'owner': None,
                 'release': '4.el7_4',
                 'resource_type': 'KojiBuild',
-                'display_name': 'slf4j-1.7.4-4.el7_4',
                 'start_time': '2017-04-02T19:39:06+00:00',
-                'state': 1,
-                'tags': [
+                'state':1,
```
mprahl: Where isn't there a space after the colon here?

tests/api/test_siblings.py
```diff
@@ -439,9 +443,329 @@ def test_node_siblings(client, resource, uid, backward_rel, expected):
     ca.attached_builds.connect(cb)
     ca.attached_builds.connect(cb_two)
 
-    url = '/api/v1/siblings/{0}/{1}'.format(resource, uid)
+    url = '/api/v1/siblings/{0}/{1}?'.format(resource, uid)
```
mprahl: Remove the question mark here.

tests/api/test_siblings.py
```diff
@@ -439,9 +443,329 @@ def test_node_siblings(client, resource, uid, backward_rel, expected):
...
+    if backward_rel:
+        url = '{0}&backward_rel=true'.format(url)
```
mprahl: Replace `&` with `?`

tests/api/test_siblings.py
```diff
@@ -439,9 +443,329 @@ def test_node_siblings(client, resource, uid, backward_rel, expected):
...
+                },
+                'version': '1.7.4'
+            }
+        ],
+        'meta': {
+            'description': 'Container builds attached to RHBA-2017:2251-03'
+        }
+    })
+])
+def test_module_story_node_siblings(client, resource, uid, backward_rel, story_path, expected):
```
mprahl: You should remove `story_path` since it's always just set to `module` since this test is only for module siblings.

tests/api/test_siblings.py
```diff
@@ -501,3 +825,31 @@ def test_reverse_flag(client):
     rv = client.get('/api/v1/siblings/containeradvisory/12327')
     assert rv.status_code == 400
     assert json.loads(rv.data.decode('utf-8')) == expected
+
+
+def test_story_path_flag(client):
+    """Tests getting the siblings for when backward_rel=false and passing in a ContainerAdvisory."""
```
mprahl: This needs updating.

tests/api/test_siblings.py
```diff
@@ -501,3 +825,31 @@ def test_reverse_flag(client):
     rv = client.get('/api/v1/siblings/containeradvisory/12327')
     assert rv.status_code == 400
     assert json.loads(rv.data.decode('utf-8')) == expected
+
+
+def test_story_path_flag(client):
```
mprahl: Change this name to `test_story_path_flag_invalid`.

estuary/api/v1.py
mprahl: To clarify, this is because we shouldn't burden Neo4j with an unnecessary query if the request is invalid.

mprahl:
@yashvardhannanavati a few minor things. Once those are addressed this should be good to go.

estuary/api/v1.py
```diff
@@ -240,14 +242,22 @@ def get_siblings(resource, uid):
     :raises NotFound: if the item is not found
     :raises ValidationError: if an invalid resource was requested
     """
+    story_type_mapper = {'container': 'ContainerStoryManager', 'module': 'ModuleStoryManager'}
+    story_type = request.args.get('story_type', 'container').lower()
+
+    story_manager_class = story_type_mapper.get(story_type)
+    if story_manager_class:
+        story_manager = getattr(estuary.utils.story, story_manager_class)()
```
mprahl: Optional: You could make the values of `story_type_mapper` be the class objects so you'd avoid having to use `getattr`.

PR #200, opened 2018-09-11
Add optional OpenID Connect authentication

mprahl:


estuary/app.py
```diff
@@ -68,11 +74,25 @@ def create_app(config_obj=None):
         load_config(app)
 
     if app.config['PRODUCTION'] and app.secret_key == 'replace-me-with-something-random':
-        raise Warning('You need to change the app.secret_key value for production')
+        raise RuntimeError('You need to change the app.secret_key value for production')
```
ralphbean: Good catch.

estuary/auth.py
```diff
@@ -0,0 +1,29 @@
...
+from flask_oidc import OpenIDConnect
+
+
+class EstuaryOIDC(OpenIDConnect):
+    """Customized version of flask_oidc.OpenIDConnect."""
+
+    def load_secrets(self, app):
+        """
+        Get the configuration required to introspect tokens using flask_oidc.
+
```
ralphbean: You might add a comment here explaining *why* you wanted to overwrite this method.

The default `load_secrets` implementation requires a second mega config file... and that seemed like overkill.  This lets us keep all the config in the same, single app config file.

estuary/utils/general.py
```diff
@@ -126,3 +129,31 @@ def get_neo4j_node(resource_name, uid):
                          '{1}, and {2}.'.format(resource_name, ', '.join(model_names[:-1]),
                                                 model_names[-1]))
                 raise ValidationError(error)
+
+
+def login_required(f):
```
ralphbean: decorators++

estuary/utils/general.py
```diff
@@ -126,3 +129,31 @@ def get_neo4j_node(resource_name, uid):
...
+    def wrapper(*args, **kwargs):
+        if current_app.config['ENABLE_AUTH']:
+            if 'Authorization' not in request.headers:
+                raise Unauthorized('An "Authorization" header wasn\'t provided')
+            token = request.headers['Authorization'].strip()
+            prefix = 'Bearer '
+            if not token.startswith(prefix):
+                raise Unauthorized(
+                    'The "Authorization" header must start with "{0}"'.format(prefix.rstrip()))
+            token = token[len(prefix):].rstrip()
```
ralphbean: This `.rstrip()` is redundant after the previous `.strip()` on line 147.

mprahl:
@ralphbean thanks for the review. I addressed your comments.

PR #203, opened 2018-09-17
Check if the user is an employee when authenticating

mprahl:


mprahl:
I agree @ralphbean. Thanks for the review. FYI, I asked in ITIAMSSO-129 to confirm that there's no way to get the scopes and the `employeeType` is being used appropriately here.

PR #207, opened 2018-10-12
Make version in docs update automatically when changed in setup.py

sarah256:


docs/conf.py
```diff
@@ -6,24 +6,31 @@
...
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#
-import os
-import sys
+
 sys.path.append(os.path.abspath('../'))
 
 # -- Project information -----------------------------------------------------
 
+log = logging.getLogger('estuary')
```
mprahl: This can be removed.

docs/conf.py
```diff
@@ -6,24 +6,31 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
+from __future__ import unicode_literals
+
+import os
+import sys
+import logging
```
mprahl: This can be removed.

setup.py
```diff
@@ -5,6 +5,7 @@
 with open('requirements.txt', 'r') as f:
     requirements = f.readlines()
 
+# Variables need to be referenced elsewhere
```
mprahl: This can be removed

docs/conf.py
sarah256: i need os and sys for the path setup part, i just moved them upwards to be with the others. the logging i will remove though

mprahl: Yeah, the comment was just on the `import logging` line.

PR #208, opened 2018-10-12
Make copyright year in docs change with the current year

sarah256:


docs/conf.py
```diff
@@ -22,12 +23,17 @@
 
 # -- Project information -----------------------------------------------------
 
+# Get current year for copyright
+t = datetime.today()
+dt = datetime(t.year, 1, 1)
+year = dt.strftime("%Y")
```
mprahl: You can simplify these three lines to:
```python
year = datetime.today().strftime('%Y')
```

docs/conf.py
```diff
@@ -22,12 +23,17 @@
...
+
 try:
     version = pkg_resources.get_distribution('estuary').version
 except pkg_resources.DistributionNotFound:
     version = 'unknown'
 project = u'Estuary API'
-copyright = u'2018, Matt Prahl, Yash Nanavati, Sarah Rieger'
+copyright = year + u', Matt Prahl, Yash Nanavati, Sarah Rieger'
```
mprahl: FYI the `u'` isn't needed since you import `from __future__ import unicode_literals`. It doesn't hurt to have them though.

sarah256: @mprahl should i delete all of the u's then?

mprahl: You don't have to. If you want to, you can do that in a separate PR. It certainly looks prettier without them. :)

sarah256: okay will do! otherwise, this should be ready to go then :smile:

PR #210, opened 2018-10-16
Update build instructions in README because makefile was removed

sarah256:
We had removed one of the make files in a previous PR, so I updated the instructions on how to build and removed the other make file


mprahl:
@sarah256 your commit message is backwards. It should be something like `Replaced the docs Makefile with simple instructions in the read me`

PR #211, opened 2018-10-19
Ignore Freshmaker dry run mode

sarah256:


scrapers/freshmaker.py
```diff
@@ -51,10 +51,13 @@ def query_api_and_update_neo4j(self):
                     # Skip Freshmaker Events that don't have the search_key as the Advisory ID
                     continue
                 log.debug('Creating FreshmakerEvent {0}'.format(fm_event['id']))
+                msg_id = fm_event['message_id']
+                if 'manual_rebuild' in msg_id:
```
mprahl: How about `if msg_id.startswith('manual_rebuild_'):`?

mprahl: Actually, it looks like `manual_rebuild_` does not necessarily mean dry run. I'll ping Jan to see if there's a way we can tell.

PR #218, opened 2018-11-14
Remove the cardinality on KojiTag.module_builds

mprahl:
Since a single module build now produces a normal build and a "-devel" build, the cardinality no longer applies.

See FACTORY-3453 for context.


yashvardhannanavati:
:+1:

PR #219, opened 2018-12-21
Allow International Local Hire to see data

gnaponie:
Allow employeeType=International Local Hire to see the data in
Estuary, since before it was allowed only to "Employee"s and it
seems that's not enough.


mprahl:
LGTM

PR #221, opened 2019-01-29
Allow monitoring users to access API endpoints that require auth

mprahl:
The IAM team isn't sure they can make a service account an employee. This a work-around for New Relic monitoring.


gnaponie:
It looks fine. Could we have some security issues because of that? I don't think so.

mprahl:
> It looks fine. Could we have some security issues because of that? I don't think so.

I think the logic is secure, but I'll only merge this if we can't get the monitoring token to be an employee.

PR #222, opened 2019-02-06
Create function to get most recent nodes from the database

sarah256:


estuary/utils/story.py
```diff
@@ -276,6 +278,42 @@ def _get_plural_label(label):
             raise RuntimeError('{0} does not have a {1} relationship'.format(
                 story_node_display_name, rel_direction))
 
+    def get_recent_nodes(self):
+        """
+        Get the most recent nodes' IDs.
```
mprahl: This needs updating

estuary/utils/story.py
```diff
@@ -276,6 +278,42 @@ def _get_plural_label(label):
             raise RuntimeError('{0} does not have a {1} relationship'.format(
                 story_node_display_name, rel_direction))
 
+    def get_recent_nodes(self):
```
mprahl: This doesn't belong in this class since it doesn't use any of the class variables or methods. It is also not part of constructing a story. It is a separate feature.

estuary/utils/story.py
```diff
@@ -276,6 +278,42 @@ def _get_plural_label(label):
...
+        """
+        Get the most recent nodes' IDs.
+
+        :param str label: Neo4j node label
```
mprahl: This needs updating

estuary/utils/story.py
```diff
@@ -276,6 +278,42 @@ def _get_plural_label(label):
...
+        :param int num_recents: the number of most recent nodes to return
```
mprahl: This needs updating

estuary/utils/story.py
```diff
@@ -276,6 +278,42 @@ def _get_plural_label(label):
...
+        :return: array of the most recent nodes
+        :rtype: array
```
mprahl: This needs updating

estuary/utils/story.py
```diff
@@ -276,6 +278,42 @@ def _get_plural_label(label):
...
+            Advisory.__label__: 'update_date'
+        }
+
+        query = 'CALL apoc.cypher.runMany(\''
+        for label, time_property in label_dict.items():
+            query += (
+                '\n    MATCH ({label}:{label}) RETURN {label} '
+                'ORDER BY {label}.{time_property} DESC LIMIT 5;'
+            ).format(label=label, time_property=time_property, rv=label.lower())
+        query = '{0}\'\n, {{}}, {{statistics: False, timeout: 15}});'.format(query)
```
mprahl: `rv` is not used, so it can be removed.

mprahl: Also, what happens when the timeout is reached when running this query from neomodel? Do we need error handling for that?

estuary/utils/story.py
```diff
@@ -276,6 +278,42 @@ def _get_plural_label(label):
...
+        log.debug('Querying Neo4j with:\n{0}'.format(query))
+
+        final_result = {}
+        results, _ = db.cypher_query(query)
+        for result in [result[1] for result in results]:
```
mprahl: Optional: A comment about why `result[1]` would be nice

tests/utils/test_story.py
```diff
@@ -33,3 +39,45 @@ def test_get_siblings_description(display_name, label, backward, expected):
     rv = story_utils.get_siblings_description(
         display_name, ContainerStoryManager().story_flow(label), backward)
     assert rv == expected
+
+
+def test_get_recent_nodes():
```
mprahl: Once you move the method this test is testing, this will need to move as well.

tests/utils/test_story.py
```diff
@@ -33,3 +39,45 @@ def test_get_siblings_description(display_name, label, backward, expected):
...
+    """Test the get_recent_nodes function."""
+    BugzillaBug.get_or_create({
+        'id_': '11111',
+        'modified_time': datetime(2017, 4, 26, 11, 44, 38)
+    })[0]
```
mprahl: `[0]` is unnecessary since you aren't storing the data in a variable. This applies for the remainder of the test.

mprahl:
This PR is missing the API endpoint to expose this to the front-end. If you'd prefer, that can be a separate PR.

estuary/utils/story.py
sarah256: where should I put it?

sarah256:
> This PR is missing the API endpoint to expose this to the front-end. If you'd prefer, that can be a separate PR.

I think I'll do this separately

estuary/utils/story.py
mprahl: You could create a new file in `estuary/utils`.

estuary/utils/story.py
sarah256: Everything else should be good, but I am unsure about this still

estuary/utils/recents.py
```diff
@@ -0,0 +1,49 @@
...
+        Advisory.__label__: 'update_date'
+    }
+
+    query = 'CALL apoc.cypher.runMany(\''
+    for label, time_property in label_dict.items():
+        query += (
+            '\n    MATCH ({label}:{label}) RETURN {label} '
+            'ORDER BY {label}.{time_property} DESC LIMIT 5;'
+        ).format(label=label, time_property=time_property)
+    query = '{0}\'\n, {{}}, {{statistics: False, timeout: 15}});'.format(query)
```
mprahl: It's not clear what the timeout actually does, so I filed an issue asking what it does:
https://github.com/neo4j-contrib/neo4j-apoc-procedures/issues/1105

PR #223, opened 2019-02-08
Add metrics for request latency and count

mprahl:


estuary/app.py
```diff
@@ -109,6 +109,14 @@ def create_app(config_obj=None):
     app.register_error_handler(ServiceUnavailable, json_error)
     app.register_error_handler(AuthError, json_error)
     app.register_blueprint(api_v1, url_prefix='/api/v1')
+    try:
+        from estuary.api.monitoring import monitoring_api, configure_monitoring
+        app.register_blueprint(monitoring_api, url_prefix='/monitoring')
+        configure_monitoring(app)
+    except ImportError as e:
```
hanzz: Not sure how it would be visible for Estuary, but some log.warning saying that /metrics endpoint is disabled because of Prometheus client not installed would be nice here.

hanzz:
The use of labels to record requests metadata is nice :).

+1 from me, the only concern is that this looks differently than what we do in other projects for monitoring, but I think this is probably not an issue... Filip probably has an opinion on that.

fivaldi:
@mprahl @hanzz It'd be nice to have it uniformed using the https://pagure.io/monitor-flask-sqlalchemy module, but it's not a crucial thing at this moment. Important is to have it in stage/prod.

fivaldi:
:+1: if the tests pass...

estuary/app.py
mprahl: Good idea. I added a warning.

mprahl:
> @mprahl @hanzz It'd be nice to have it uniformed using the https://pagure.io/monitor-flask-sqlalchemy module, but it's not a crucial thing at this moment. Important is to have it in stage/prod.

I would tend to agree, but Estuary doesn't use SQLAlchemy. It uses neomodel as its OGM since it doesn't use a relational database.

PR #225, opened 2019-02-18
Add API endpoint for recent stories

sarah256:


tests/api/test_recents.py
```diff
@@ -0,0 +1,153 @@
...
+import pytest
+
+from estuary.models.koji import KojiBuild
+from estuary.models.bugzilla import BugzillaBug
+from estuary.models.distgit import DistGitCommit
+from estuary.models.errata import Advisory
+from estuary.models.freshmaker import FreshmakerEvent
+
+
+@pytest.mark.parametrize('expected', [
```
mprahl: Since you're only doing one test, `parametrize` isn't needed. Can you refactor the test to not use it?

PR #227, opened 2019-02-22
Increase the request timeout to Freshmaker's API in the Freshmaker scraper

mprahl:
Some API calls take longer than 15 seconds, so it's best to just increase the timeout.


gnaponie:
+1

PR #228, opened 2019-02-25
Rewrite cypher query so that it goes through each node type individually instead of as one large query

sarah256:


estuary/utils/recents.py
```diff
@@ -29,21 +28,17 @@ def get_recent_nodes():
         Advisory.__label__: 'update_date'
     }
 
-    query = 'CALL apoc.cypher.runMany(\''
+    final_result = {}
     for label, time_property in label_dict.items():
-        query += (
-            '\n    MATCH ({label}:{label}) RETURN {label} '
+        query = (
+            'MATCH ({label}:{label}) RETURN {label} '
```
mprahl: Can you call the variable name something generic? It's a bit confusing that `label` is used everywhere. We did that before because `runMany` needed distinct variables for each query, but we don't need to anymore. You could do something like `item` instead.

sarah256: Just to confirm, change all three to `item` ?

estuary/utils/recents.py
```diff
@@ -29,21 +28,17 @@ def get_recent_nodes():
...
-    query = '{0}\'\n, {{}}, {{statistics: False, timeout: 15}});'.format(query)
-    log.debug('Querying Neo4j with:\n{0}'.format(query))
-
     final_result = {}
-    results, _ = db.cypher_query(query)
-    for result in [result[1] for result in results]:
-        for label, raw_node in result.items():
+    for label, time_property in label_dict.items():
+        query = (
+            'MATCH ({item}:{item}) RETURN {item} '
```
mprahl: It should be something like this:
```python
query = (
    'MATCH (node:{labe}) RETURN node '
    'ORDER BY node.{time_property} DESC LIMIT 5'
).format(label=label, time_property=time_property) 
```

PR #234, opened 2019-04-08
Display failed and in-progress container builds

gnaponie:
We currently just display successful builds.


gnaponie:
Related PRs:
https://github.com/release-engineering/estuary/pull/83
https://github.com/release-engineering/estuary-updater/pull/64

mprahl:
@gnaponie could you please fix the flake8 errors?

estuary/models/freshmaker.py
```diff
@@ -20,9 +20,36 @@ class FreshmakerEvent(EstuaryStructuredNode):
...
         return 'Freshmaker event {0}'.format(self.id_)
+
+
+class FreshmakerBuild(EstuaryStructuredNode):
+    """Definition of a Freshmaker build in Neo4j."""
+
+    id_ = UniqueIdProperty(db_property='id')
+    build_id = IntegerProperty(unique=True, required=True)
+    dep_on = StringProperty()
+    event_id = IntegerProperty(required=True)
```
mprahl: Instead of storing the `event_id`, maybe it makes more sense to have a relationship of:
```python
event = RelationshipFrom('.FreshmakerEvent, 'TRIGGERED')
```

estuary/models/freshmaker.py
```diff
@@ -20,9 +20,36 @@ class FreshmakerEvent(EstuaryStructuredNode):
...
     def display_name(self):
         """Get intuitive (human readable) display name for the node."""
         return 'Freshmaker event {0}'.format(self.id_)
+
+
+class FreshmakerBuild(EstuaryStructuredNode):
+    """Definition of a Freshmaker build in Neo4j."""
+
+    id_ = UniqueIdProperty(db_property='id')
+    build_id = IntegerProperty(unique=True, required=True)
```
mprahl: My preference it to remove `required` to all of these properties because it affects the usability of `get_or_create`. For instance, if I need to add a relationship to a `FreshmakerBuild` node and all I have is the ID, normally I'd first start by either getting the build or creating it with `FreshmakerBuild.get_or_create({'id_': 12345})`, but if all those other properties are required, then `get_or_create` will require them also.

All the required properties also determine the uniqueness of a node in `get_or_create`, so if for some reason the `url` property changes, and I run `get_or_create` with all the same information except for the URL being different, an exception will be raised saying that a new `FreshmakerBuild` was being created, but it violates the unique constraint on `id_`.

estuary/models/koji.py
```diff
@@ -66,7 +66,7 @@ class ContainerKojiBuild(KojiBuild):
     """A Neo4j definition of a build that represents a container build in Koji."""
 
     original_nvr = StringProperty()
-    triggered_by_freshmaker_event = RelationshipFrom(
```
mprahl: I prefer this original name. What was the motivation for changing it?

scrapers/freshmaker.py
```diff
@@ -70,8 +70,30 @@ def query_api_and_update_neo4j(self):
 
                 for build_dict in fm_event['builds']:
                     # To handle a faulty container build in Freshmaker
-                    if not build_dict['build_id'] or int(build_dict['build_id']) < 0:
```
mprahl: Why was this removed? There was a Freshmaker bug at one point that caused the IDs to be negative and I think those are still presented in the API. If that's the case, then the scraper still needs to be able to ignore them.

scrapers/freshmaker.py
```diff
@@ -70,8 +70,30 @@ def query_api_and_update_neo4j(self):
...
+                        state=build_dict['state'],
+                        state_name=build_dict['state_name'],
+                        state_reason=build_dict['state_reason'],
+                        time_submitted=timestamp_to_datetime(build_dict['time_submitted']),
+                        type_=build_dict['type'],
+                        type_name=build_dict['type_name'],
+                        url=build_dict['url']
+                    )
+                    if build_dict['time_completed']:
+                        cb_params['time_completed'] = timestamp_to_datetime(
```
mprahl: `cb_params` => `fb_params`

mprahl:
I left a few comments that need to be addressed, but other than that, it looks great!

estuary/models/koji.py
gnaponie: I thought it was more representative. I'll change it back.

scrapers/freshmaker.py
gnaponie: I'll put it back. I thought the negative build id meant failed build. The comment says "faulty container build in Freshmaker", is it related only to the first condition of the if?

gnaponie:
@mprahl I think I've addressed all the comments. Could you check again? Thank you

PR #236, opened 2019-04-29
Add metadata for recents endpoint that specifies the id and timestamp keys

sarah256:


sarah256:
@mprahl

estuary/api/v1.py
```diff
@@ -7,6 +7,11 @@
 
 from estuary import version
 from estuary.models.base import EstuaryStructuredNode
+from estuary.models.koji import KojiBuild
```
mprahl: It looks like these imports aren't needed

estuary/api/v1.py
```diff
@@ -348,8 +353,7 @@ def get_recent_stories():
     """Get stories that were most recently updated, by their artifact type."""
     nodes = get_recent_nodes()
```
mprahl: Can you make this:
```python
nodes, meta = get_recent_nodes()
```

estuary/utils/recents.py
```diff
@@ -16,20 +16,32 @@ def get_recent_nodes():
...
     """
-    label_dict = {
+    timestamp_dict = {
         FreshmakerEvent.__label__: 'id',
         BugzillaBug.__label__: 'modified_time',
         DistGitCommit.__label__: 'commit_date',
         KojiBuild.__label__: 'completion_time',
         Advisory.__label__: 'update_date'
     }
+    id_dict = {
```
mprahl: This should just be an empty dictionary since your code below will create these entries even if they don't already exist.

estuary/utils/recents.py
```diff
@@ -38,9 +50,10 @@ def get_recent_nodes():
         ).format(label=label, time_property=time_property)
         results, _ = db.cypher_query(query)
         for result in results:
-            node_results = final_result.setdefault(label, [])
+            node_results = final_result_data.setdefault(label, [])
             # result is always a list of a single node
             node = inflate_node(result[0])
             node_results.append(node.serialized)
+            id_dict[node.__label__] = node.unique_id_property
```
mprahl: As an efficiency improvement, you could check if the `id_dict[node.__label__]` is already set before setting it since `unique_id_property` is a property method that executes code to determine the value, and it's a bit wasteful to keep running that over and over again.

tests/utils/test_recents.py
```diff
@@ -47,10 +47,28 @@ def test_get_recent_nodes():
 
     ret = estuary.utils.recents.get_recent_nodes()
```
mprahl: Can you change `ret` to `nodes, meta`. It'd make the code below easier to read.

PR #239, opened 2019-05-13
Fix a typo in the Freshmaker scraper

mprahl:
In the previous PR, I had made a mistake. This line that I'm deleting has a typo in it, but it's not actually needed because of the line below it.


gnaponie:
+1!

PR #241, opened 2019-05-21
Update container base images from Fedora 28 to Fedora 30

sarah256:


mprahl:
@sarah256 TravisCI is failing. The issue is with the following in the Dockerfile:
```
# We must install prometheus_client for Python 2 since it is no longer packaged for Fedora
RUN pip install flake8-docstrings prometheus_client
```

I think we'll need to use `pip3` instead. Also you can remove `prometheus_client` from the `pip install` based on the comment in the Dockerfile.

docker/Dockerfile-tests
```diff
@@ -1,32 +1,24 @@
...
   python3-pytest-cov \
   && dnf clean all
-# We must install prometheus_client for Python 2 since it is no longer packaged for Fedora
-RUN pip install flake8-docstrings prometheus_client
+RUN pip3 install flake8-docstrings 
 
 VOLUME /src
 WORKDIR /src
 # Inspired from https://github.com/neo4j-contrib/neomodel/blob/master/tests-with-docker-compose.sh
-CMD ["sh", "-c", "while ! nc -z -w 2 neo4j 7687; do sleep 1; done; pytest-2 -vvv --cov-report term-missing --cov=estuary tests/ && pytest-3 -vvv --cov-report term-missing --cov=estuary tests/ && flake8"]
```
sarah256: removing the pytest-2 portion triggers a socket error

mprahl:
@sarah256 this looks good. It seems we also need to the URL encoding from the frontend as well since providing "RHBZ#12345" doesn't currently work from the frontend. Could you please file a ticket to fix this? After that, let's merge this.

mprahl:
@sarah256 the commit history looks a little wonky. Could you please fix it?

gnaponie:
Sorry, I didn't see it until now :(
Anyhow, after-merge +1 \o/

PR #243, opened 2019-06-03
Add label and display name to recent nodes

sarah256:


estuary/utils/recents.py
```diff
@@ -46,7 +46,10 @@ def get_recent_nodes():
             node_results = final_result_data.setdefault(label, [])
             # result is always a list of a single node
             node = inflate_node(result[0])
-            node_results.append(node.serialized)
+            serialized_node = node.serialized
+            serialized_node['resource_type'] = node.__label__
+            serialized_node['display_name'] = node.display_name
```
gnaponie: In case these fields are "None", are we just gonna show "None"? How was it before? Is there maybe a way to show a more useful information?

sarah256: I think for the label they will never be none because it is just based off of what type of node/artifact it is. As for the display name, I'm not sure. @mprahl thoughts?

mprahl: Yes, each model class that this API endpoint will use implements a property method of `display_name`. If they don't, a `NotImplementedError` exception is raised from the base class.

tests/api/test_recents.py
```diff
@@ -45,15 +45,19 @@ def test_get_recent_nodes(client):
                     'status_time': None,
                     'actual_ship_date': None,
                     'release_date': None,
-                    'id': '66666'
+                    'id': '66666',
+                    'display_name': None,
```
mprahl: For this not to be `None`, perhaps you can add a value to `advisory_name`.

tests/api/test_recents.py
```diff
@@ -77,7 +83,9 @@ def test_get_recent_nodes(client):
                     'epoch': None,
                     'version': None,
                     'release': None,
-                    'id': '44444'
+                    'id': '44444',
+                    'display_name': 'None-None-None',
```
mprahl: Can you add `name`, `version`, and `release` to the `KojiBuild` you create so it's not `None-None-None`?

PR #246, opened 2019-06-21
Update Freshmaker models to include time_created and time_done

sarah256:


sarah256:
closed due to git mistake, see #247 for the updated PR

PR #247, opened 2019-06-21
Update Freshmaker scrapers and models to include time_created and time_done

sarah256:


scrapers/freshmaker.py
```diff
@@ -62,15 +62,20 @@ def query_api_and_update_neo4j(self):
...
                     id_=fm_event['id'],
                     event_type_id=fm_event['event_type_id'],
                     message_id=fm_event['message_id'],
                     state=fm_event['state'],
                     state_name=fm_event['state_name'],
                     state_reason=fm_event['state_reason'],
                     url=fm_event['url']
-                ))[0]
+                )
+                if fm_event['time_created']:
```
mprahl: This will raise a key error if it's not present in the event. How about:
```python
if 'time_created' in fm_event:
```

scrapers/freshmaker.py
```diff
@@ -62,15 +62,20 @@ def query_api_and_update_neo4j(self):
...
+                    event_params['time_created'] = fm_event['time_created']
+                if fm_event['time_done']:
```
mprahl: Same as above

PR #248, opened 2019-06-21
Expose timeline date for all artifacts in a story in a consistent format

sarah256:


mprahl:
@sarah256 could you actually have those dates be formatted in the ISO 8601 format? That is more conventional for a REST API to return, and it will allow your JavaScript code to convert it to a Date object. This can then allow you to display it in any format or timezone that you'd like. Here is an example of that in another project:
https://github.com/release-engineering/mbs-ui/blob/master/src/app/mbs/module-detail/module-detail.component.html#L98

sarah256:
@mprahl This should be ready for review again!

estuary/api/v1.py
```diff
@@ -337,6 +339,7 @@ def get_artifact_relationships(resource, uid, relationship):
         serialized_node = node.serialized_all
         serialized_node['resource_type'] = node.__label__
         serialized_node['display_name'] = node.display_name
+        serialized_node['timeline_property'] = node.timeline_property
```
mprahl: Depending on the `relationship` that is requested, this will raise a `NotImplementedError`. One example would be requesting `/api/v1/relationships/freshmakerevent/1234/requested_builds`. This line could probably just be deleted since it's not important for this API endpoint.

PR #250, opened 2019-06-24
Switch to use ISO format with timezone in serialized method

sarah256:


mprahl:
:+1:

estuary/models/base.py
```diff
@@ -35,7 +35,7 @@ def serialized(self):
             actual_key = getattr(self.__class__, key).db_property or key
 
             if isinstance(value, datetime):
-                rv[actual_key] = value.isoformat()
+                rv[actual_key] = value.strftime("%Y-%m-%dT%H:%M:%SZ")
```
mprahl: Could you please use single quotes since the rest of code base uses them?

PR #252, opened 2019-06-25
Add a function to calculate wait time between each artifact and expose it to the metadata in the REST API

sarah256:


tests/api/test_all_stories.py
```diff
@@ -228,6 +229,8 @@
             'meta': {
                 'story_related_nodes_backward': [0, 0, 0, 1, 0, 1],
                 'story_related_nodes_forward': [0, 1, 0, 0, 0, 0],
+                'wait_times': [[23, 16, 5], [-24, 7, 54], [0, 19, 8], [869, 22, 54],
+                               [-29726, 5, 56]],
```
yashvardhannanavati: I wonder why there are negative values in the `wait_times`. Am I missing something?

sarah256: @yashvardhannanavati It is because the test dates are not accurate to what the real pipeline would look like. I did the math for each and they check out, but since the fake data is not in the real order it returns weird numbers like this.

estuary/utils/story.py
```diff
@@ -144,6 +145,51 @@ def get_sibling_nodes_count(self, results, reverse=False):
...
+        for index in range(len_story - 1):
+            completion_time = getattr(results[index], completion_times[results[index].__label__])
+            start_time = datetime.strptime(results[index + 1].timeline_timestamp,
+                                           '%Y-%m-%dT%H:%M:%SZ')
+
+            # Remove timezone info so that both are offset naive and thus able to be subtracted
+            start_time = start_time.replace(tzinfo=None)
+            completion_time = completion_time.replace(tzinfo=None)
+
+            wait_time = start_time - completion_time
```
yashvardhannanavati: should this be `completion_time - start_time` instead? Because `completion_time` should and will always be bigger right?

tests/api/test_all_stories.py
yashvardhannanavati: Is it because the wait_time is calculated using `start_time - completion_time`?

estuary/utils/story.py
```diff
@@ -144,6 +145,51 @@ def get_sibling_nodes_count(self, results, reverse=False):
             return correlated_nodes[::-1]
         return correlated_nodes
 
+    def get_wait_times(self, results):
+        """
+        Get the wait time between two different artifacts.
+
+        :param list results: contains inflated results from Neo4j
+        :return: list of wait time tuples in order of the pipeline (oldest to newest)
+        :rtype: list
```
mprahl: Could you add the following to the docstring (not sure if it's `raise` or `raises`)?
```python
:raises RuntimeError: if results has less than 2 elements
```

estuary/utils/story.py
sarah256: @yashvardhannanavati I see what you're saying, but for here this is written this way because it is the start time of the next node compared with the completion of the current one, and we want the wait time to be the time from the current one finishing until the next one starts. Lmk if that doesn;t make sense still.

estuary/utils/story.py
```diff
@@ -144,6 +145,51 @@ def get_sibling_nodes_count(self, results, reverse=False):
...
+            'FreshmakerEvent': 'time_done',
+            'KojiBuild': 'completion_time',
+            'ModuleKojiBuild': 'completion_time',
+            'ContainerKojiBuild': 'completion_time'
+        }
+
+        wait_times = []
+
+        for index in range(len_story - 1):
+            completion_time = getattr(results[index], completion_times[results[index].__label__])
```
mprahl: Optional: It'd be easier to read if you did:
```python
property_name = completion_times[results[index].__label__]
completion_time = getattr(results[index], property_name)
```

mprahl: Could you assign `results[index]` to a variable like `artifact` for readability?

estuary/utils/story.py
yashvardhannanavati: makes sense. Thanks!

estuary/utils/story.py
yashvardhannanavati: Or a comment like `completion_time of the current artifact` on this line and `start_time of the next artifact in the story` on the next would do the trick too. Wdyt?

tests/api/test_all_stories.py
yashvardhannanavati: Ah, makes sense. It would be great to fix a few just to be sure. :)

estuary/utils/story.py
sarah256: @yashvardhannanavati is this okay:

# Find the time between when the current artifact completes and the next one starts
 wait_time = start_time - completion_time

sarah256: oops didnt realize a `#` made it bold lol that's supposed to be a comment

estuary/utils/story.py
```diff
@@ -144,6 +145,51 @@ def get_sibling_nodes_count(self, results, reverse=False):
...
+            start_time = datetime.strptime(results[index + 1].timeline_timestamp,
```
mprahl: Could you assign `results[index + 1]` to a variable like `next_artifact` for readability?

mprahl: I think you'll need to make sure that `completion_time` and `results[index + 1].timeline_timestamp` are not `None`

mprahl: Could you also rename `start_time` to `next_artifact_start_time` or something like that. It makes it easier to read.

tests/api/test_all_stories.py
mprahl: @sarah256 would it make sense to check `wait_time < timedelta()`? If it is, then it's an invalid timedelta and an exception can be raised. Any code that calls that function would need to ignore it though.

mprahl: Also, to @yashvardhannanavati's point, could you please add a test or edit an existing one that has valid timestamps?

gnaponie:
The PR looks fine. I agree with Matt and Yash for the comments.
Especially it would be good to have at least one test that covers the case without negative values.

sarah256:
> The PR looks fine. I agree with Matt and Yash for the comments.
> Especially it would be good to have at least one test that covers the case without negative values.

Thanks! I just changed it so that there is some proper examples in `tests/api/test_all_stories.py`

estuary/utils/story.py
```diff
@@ -144,6 +145,59 @@ def get_sibling_nodes_count(self, results, reverse=False):
...
+        }
+
+        wait_times = []
+
+        for index in range(len_story - 1):
+            artifact = results[index]
+            next_artifact = results[index + 1]
+            property_name = completion_times[artifact.__label__]
+            if getattr(artifact, property_name) and next_artifact.timeline_timestamp:
+                completion_time = getattr(artifact, property_name)
```
mprahl: What if `completion_time` is `None` if Estuary-Updater hasn't filled in this info yet?

estuary/utils/story.py
```diff
@@ -144,6 +145,59 @@ def get_sibling_nodes_count(self, results, reverse=False):
             return correlated_nodes[::-1]
         return correlated_nodes
 
+    def get_wait_times(self, results):
```
mprahl: Could you add a specific unit test for this method?

mprahl: After discussion, we decided that fixing the timestamps in `test_stories.py` would be sufficient.

estuary/utils/story.py
```diff
@@ -144,6 +145,59 @@ def get_sibling_nodes_count(self, results, reverse=False):
...
+            if not completion_time or not next_artifact.timeline_timestamp:
+                continue
+
+            next_artifact_start_time = datetime.strptime(next_artifact.timeline_timestamp,
+                                                         '%Y-%m-%dT%H:%M:%SZ')
+            # Remove timezone info so that both are offset naive and thus able to be subtracted
+            next_artifact_start_time = next_artifact_start_time.replace(tzinfo=None)
+            completion_time = completion_time.replace(tzinfo=None)
+
+            if completion_time > next_artifact_start_time:
```
mprahl: Optional: Adding a comment explaining this would be nice

sarah256:
NOTE: The reason the Travis failed is because the PR was merged before it finished. The only addition, which triggered the new build, was a comment, and the build worked prior.

PR #253, opened 2019-06-26
Show all identified images

gnaponie:
A recent change introduced the chance to see also failed Freshmaker
builds and not only the successful. Some of them seem to be skipped
due to failure building parent images. With this change they will
also be showed.
ref: FACTORY-4592

Signed-off-by: gnaponie <gnaponie@redhat.com>

### DIFF
--- a/scrapers/freshmaker.py
+++ b/scrapers/freshmaker.py
@@ -86,12 +86,11 @@ class FreshmakerScraper(BaseScraper):
 
                 for build_dict in fm_event['builds']:
                     # To handle a faulty container build in Freshmaker
-                    if not build_dict['build_id'] or int(build_dict['build_id']) < 0:
+                    if build_dict['build_id'] and int(build_dict['build_id']) < 0:
                         continue
                     log.debug('Creating FreshmakerBuild {0}'.format(build_dict['build_id']))
                     fb_params = dict(
                         id_=build_dict['id'],
-                        build_id=build_dict['build_id'],
                         dep_on=build_dict['dep_on'],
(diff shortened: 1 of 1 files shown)


gnaponie:
Related PR: https://github.com/release-engineering/estuary-updater/pull/68

scrapers/freshmaker.py
```diff
@@ -86,12 +86,11 @@ def query_api_and_update_neo4j(self):
 
                 for build_dict in fm_event['builds']:
                     # To handle a faulty container build in Freshmaker
-                    if not build_dict['build_id'] or int(build_dict['build_id']) < 0:
+                    if int(build_dict['build_id']) < 0:
```
mprahl: Could you please explain this change? This code was protecting against `build_dict['build_id']` was `None`. You can see that the model doesn't have `nullable=False` set:
https://pagure.io/freshmaker/blob/master/f/freshmaker/models.py#_501

gnaponie: Yeah I see. But the request is exactly to show in Estuary those builds... Example: https://pipeline.engineering.redhat.com/freshmakerevent/13277

Can we change the constrict in the model?

mprahl: I think changing this line to the following would alleviate my concerns:
```python
if build_dict['build_id'] and int(build_dict['build_id']) < 0:
```

gnaponie: Sounds good! Thank you, Matt.
I've addressed the comment.

PR #254, opened 2019-06-27
Functions to calculate total wait, processing, and lead times in the pipeline

sarah256:


estuary/api/v1.py
```diff
@@ -121,6 +121,9 @@ def _get_partial_story(results, reverse=False):
         rv['meta']['requested_node_index'] = 0
         rv['meta']['story_type'] = story_manager.__class__.__name__[:-12].lower()
         rv['meta']['wait_times'] = [0]
+        rv['meta']['total_wait_time'] = [0]
+        rv['meta']['total_processing_time'] = [0]
```
mprahl: Optional: Wouldn't this be available to calculate?

estuary/api/v1.py
```diff
@@ -233,6 +236,9 @@ def _get_partial_stories(results, reverse=False):
...
```
mprahl: Optional: Wouldn't this be available to calculate?

estuary/api/v1.py
```diff
@@ -233,6 +236,9 @@ def _get_partial_stories(results, reverse=False):
...
+        rv['meta']['total_lead_time'] = [0]
```
mprahl: Optional: Couldn't you calculate this?

estuary/api/v1.py
```diff
@@ -121,6 +121,9 @@ def _get_partial_story(results, reverse=False):
...
```
mprahl: Optional: Couldn't you calculate this?

estuary/utils/story.py
```diff
@@ -150,8 +150,9 @@ def get_wait_times(self, results):
         Get the wait time between two different artifacts.
```
mprahl: This doc string is a bit misleading. Could you clarify that this is for the whole story?

estuary/utils/story.py
```diff
@@ -150,8 +150,9 @@ def get_wait_times(self, results):
...
 
         :param list results: contains inflated results from Neo4j
-        :return: list of wait time ints in order of the pipeline (oldest to newest)
-        :rtype: list
+        :return: tuple with list of wait time ints in order of the pipeline (oldest to newest), and
```
mprahl: Can you use the word `story` instead of `pipeline`?

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
 
             # Find the time between when the current artifact completes and the next one starts
             wait_time = next_artifact_start_time - completion_time
-            wait_times[index] = wait_time.seconds
+            wait_times[index] = wait_time.seconds + (wait_time.days * 86400)
```
mprahl: Would `wait_time.total_seconds` work instead?

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+
+            # The 'wait time' between a FreshmakerEvent and a ContainerKojiBuild is still a part of
+            # the processing in a FreshmakerEvent, so we do not count it towards the total wait time
+            if artifact.__label__ != 'FreshmakerEvent':
+                total_wait_time += wait_time.seconds + (wait_time.days * 86400)
```
mprahl: Same as above

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
 
-        return wait_times
+        return (wait_times, total_wait_time)
```
mprahl: Optional: You can omit the parentheses

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+
+    def get_total_processing_time(self, results):
+        """
+        Get the total time spent processing in the pipeline.
```
mprahl: Can you use the word `story` instead of `pipeline`?

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+
+        :param list results: contains inflated results from Neo4j
+        :return: the seconds of total time spent processing
+        :rtype: int
+        """
+        len_story = len(results)
+        if len_story < 2:
+            raise RuntimeError('This function can\'t be called with one or zero elements')
```
mprahl: I don't think this restriction is necessary. The processing time can be calculated on a single node.

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+        total = 0
+        timed_processes = {
+            'FreshmakerEvent': ['time_created', 'time_done'],
+            'KojiBuild': ['creation_time', 'completion_time'],
+            'ModuleKojiBuild': ['creation_time', 'completion_time'],
+            'ContainerKojiBuild': ['creation_time', 'completion_time']
+        }
+        for index in range(len_story - 1):
+            artifact = results[index]
+            if artifact.__label__ in timed_processes.keys():
```
mprahl: `.keys()` is not necessary.

Also, it might be easier to read if you did:

```python
if artifact.__label__ not in timed_processes:
    continue

# We do not want the processing time of the entire FreshmakerEvent, just the
....
```

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+                # We do not want the processing time of the entire FreshmakerEvent, just the
+                # processing time until the displayed ContainerKojiBuild is complete
+                if artifact.__label__ == 'FreshmakerEvent' and results[index + 1]:
```
mprahl: Why are you checking `results[index + 1]`? Won't that always be truthy?

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+                    creation_time = getattr(artifact, timed_processes['FreshmakerEvent'][0])
+                    completion_time = getattr(results[index + 1],
+                                              timed_processes['ContainerKojiBuild'][1])
```
mprahl: Could you use `results[index + 1].__label__` instead? Bonus points if you assign `results[index + 1]` to a variable  :)

mprahl: I think we decided that the process time of a `FreshmakerEvent` is the start time of the `ContainerKojiBuild` minus the start time of the `FreshmakerEvent`, not the completion time.

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+                # We have already accomodated for the ContainerKojiBuild in the previous clause
+                elif artifact.__label__ == 'ContainerKojiBuild':
```
mprahl: Based on my previous comment, this `elif` can be removed.

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+                    continue
+                else:
+                    creation_time = getattr(artifact, timed_processes[artifact.__label__][0])
+                    completion_time = getattr(artifact, timed_processes[artifact.__label__][1])
+                processing_time = completion_time - creation_time
+                total += processing_time.seconds + (processing_time.days * 86400)
```
mprahl: Can you use `processing_time.total_seconds()` instead?

estuary/utils/story.py
```diff
@@ -260,6 +304,9 @@ def format_story_results(self, results, requested_item):
             serialized_node['display_name'] = node.display_name
             serialized_node['timeline_timestamp'] = node.timeline_timestamp
             data.append(serialized_node)
+
+        wait_times = self.get_wait_times(results)
```
mprahl: How about using two variables for readability?

```python
 wait_times, total_wait_time = self.get_wait_times(results)
```

estuary/utils/story.py
```diff
@@ -195,9 +197,51 @@ def get_wait_times(self, results):
...
+        if len_story < 2:
+            raise RuntimeError('This function can\'t be called with one or zero elements')
+        total = 0
+        timed_processes = {
+            'FreshmakerEvent': ['time_created', 'time_done'],
+            'KojiBuild': ['creation_time', 'completion_time'],
+            'ModuleKojiBuild': ['creation_time', 'completion_time'],
+            'ContainerKojiBuild': ['creation_time', 'completion_time']
+        }
+        for index in range(len_story - 1):
```
mprahl: What about the last node?

estuary/utils/story.py
sarah256: To see if there is a container build after it. Will that always be the case? I thought it was possible that there could be an event that hadn't triggered any builds yet.

estuary/utils/story.py
sarah256: If I assign results[index+1] to a variable before it is used, couldn't it potentially lead to an out of bounds error?

sarah256: With the process time, doing this includes both the processing times of the event and the build. Since we don't need them separately, the total processing time will be the same. If the problem is with readability though, will happily change it.

estuary/api/v1.py
sarah256: What would be passed in for `results`? Since `results` must be nothing in order to get to this part of the code.

mprahl:
@sarah256 let me know when this is ready for review again.

sarah256:
@mprahl This should be good now! Sorry for the delay - I had noticed some edge case issues when I was working on the front end and forgot to push them all.

estuary/api/v1.py
mprahl: @sarah256 where did we land on this?

estuary/models/errata.py
```diff
@@ -53,6 +53,15 @@ def timeline_timestamp(self):
             return self.created_at.strftime('%Y-%m-%dT%H:%M:%SZ')
         return None
 
+    @property
+    def attached_build_times(self):
```
mprahl: Based on how you use this, it seems wasteful to query for all the attached builds just to see if the specific build in the story is attached and has a build time.

How about replacing this with:
```python
rel = advisory.attached_builds.relationship(build)
time_attached = getattr(rel, 'time_attached', None)
```

estuary/utils/story.py
```diff
@@ -185,6 +188,13 @@ def get_wait_times(self, results):
 
             next_artifact_start_time = datetime.strptime(next_artifact.timeline_timestamp,
                                                          '%Y-%m-%dT%H:%M:%SZ')
+
+            if next_artifact.__label__ in ['Advisory', 'ContainerAdvisory'] and \
+                    artifact.__label__ in ['KojiBuild', 'ContainerKojiBuild', 'ModuleKojiBuild']:
+                id_num = getattr(artifact, 'id_')
```
mprahl: You don't need to use `getattr`. `id_num = artifact.id_` works just fine.

estuary/utils/story.py
```diff
@@ -185,6 +188,13 @@ def get_wait_times(self, results):
 
             next_artifact_start_time = datetime.strptime(next_artifact.timeline_timestamp,
                                                          '%Y-%m-%dT%H:%M:%SZ')
+
+            if next_artifact.__label__ in ['Advisory', 'ContainerAdvisory'] and \
+                    artifact.__label__ in ['KojiBuild', 'ContainerKojiBuild', 'ModuleKojiBuild']:
```
mprahl: I don't think this "and" is necessary. An advisory is always on its own or after these labels.

estuary/utils/story.py
```diff
@@ -185,6 +188,13 @@ def get_wait_times(self, results):
 
             next_artifact_start_time = datetime.strptime(next_artifact.timeline_timestamp,
```
mprahl: Optional: If you put this in the else of the if statement below, you avoid needing to parse the timestamp.

mprahl: Optional: It seems a bit clunky to convert a `datetime` object to a string in the `timeline_timestamp` property method and then converting right back to a `datetime` object. It might be cleaner to have a second property method that returns a `datetime` object and then `timeline_timestamp` calls that new property method, or you can just have `timeline_timestamp` return a `datetime` object.

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+            # The 'wait time' between a FreshmakerEvent and a ContainerKojiBuild is still a part of
+            # the processing in a FreshmakerEvent, so we do not count it towards the total wait time
+            if artifact.__label__ != 'FreshmakerEvent':
+                total_wait_time += wait_time.total_seconds()
+
+        return wait_times, total_wait_time
+
+    def get_total_processing_time(self, results):
+        """
+        Get the total time spent processing in the story.
```
mprahl: `processing in the` => `processing the`

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+        }
+        for index, artifact in enumerate(results):
+            if artifact.__label__ not in timed_processes:
+                continue
+
+            creation_time = getattr(artifact, timed_processes[artifact.__label__][0])
+            if not creation_time:
+                id_num = getattr(artifact, artifact.unique_id_property + '_')
+                log.warning(
+                    'While calculating the total processing time, a %s with id %s was encountered '
```
mprahl: Optional: `id` => `ID`

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+                    'without a creation time.',
+                    artifact.__label__, id_num
+                )
+                continue
```
mprahl: I'm wondering if the method should return `None` in this case since the total processing time will be inaccurate. Let's leave it for now, but could you reach out to Ralph about this?

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+
+            if artifact.__label__ in ['KojiBuild', 'ModuleKojiBuild', 'ContainerKojiBuild']:
+                build = artifact
+
+            if artifact.__label__ in ['Advisory', 'ContainerAdvisory']:
```
mprahl: Could you make this an `elif`?

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+            if not creation_time:
+                id_num = getattr(artifact, artifact.unique_id_property + '_')
+                log.warning(
+                    'While calculating the total processing time, a %s with id %s was encountered '
+                    'without a creation time.',
+                    artifact.__label__, id_num
+                )
+                continue
+
+            if artifact.__label__ in ['KojiBuild', 'ModuleKojiBuild', 'ContainerKojiBuild']:
```
mprahl: Optional: You could just do
```python
if artifact.__label__.endswith('KojiBuild'):
```

estuary/utils/story.py
mprahl: Optional: You could just do:
```python
if artifact.__label__.endswith('Advisory'):
```

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+                build = artifact
+
+            if artifact.__label__ in ['Advisory', 'ContainerAdvisory']:
+                if artifact.state not in ['SHIPPED_LIVE', 'DROPPED_NO_SHIP']:
```
mprahl: Optional: When doing an if/else, it's easier to read when you don't `not`. For example:

```python
if artifact.state in ['SHIPPED_LIVE', 'DROPPED_NO_SHIP']:
    completion_time = getattr(artifact, timed_processes[artifact.__label__][1])
else:
    completion_time = datetime.utcnow()
```

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+                    completion_time = datetime.utcnow()
+                else:
+                    completion_time = getattr(artifact, timed_processes[artifact.__label__][1])
+                if build:
+                    id_num = getattr(build, 'id_')
```
mprahl: `getattr` is not required here. `id_name = build.id_` works just fine.

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
+
+            # We do not want the processing time of the entire FreshmakerEvent, just the
+            # processing time until the displayed ContainerKojiBuild is created
+            elif artifact.__label__ == 'FreshmakerEvent':
+                if index != len(results) - 1:
+                    next_artifact = results[index + 1]
+                    completion_time = getattr(next_artifact,
+                                              timed_processes[next_artifact.__label__][0])
+                elif artifact.state_name in ['COMPLETE', 'SKIPPED', 'FAILED', 'CANCELED']:
+                    completion_time = getattr(artifact, timed_processes['FreshmakerEvent'][1])
```
mprahl: What if this is a legacy `FreshmakerEvent` where `time_done` is `None`?

estuary/utils/story.py
```diff
@@ -195,9 +205,133 @@ def get_wait_times(self, results):
...
-        return wait_times
+            else:
+                completion_time = getattr(artifact, timed_processes[artifact.__label__][1])
+                if not completion_time:
+                    completion_time = datetime.utcnow()
+
+            # Remove timezone info so that both are offset naive and thus able to be subtracted
+            creation_time = creation_time.replace(tzinfo=None)
+            completion_time = completion_time.replace(tzinfo=None)
+            processing_time = completion_time - creation_time
```
mprahl: Optional: What do you think about adding a warning log message if the processing time is less than 0? Then we could go through the logs to find any bugs or incorrect data.

estuary/models/errata.py
sarah256: Doing this was more for the front end so that it'd be easier to get a creation time. I'm not sure that I could get it easily elsewhere.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 Flask
 neomodel>=3.2.8
 six
+pytz
```
mprahl: Could you alphabetize this?

tests/api/test_timeline.py
```diff
@@ -0,0 +1,324 @@
...
+
+    assert processing_time == expected[0]
+    assert lead_time == expected[1]
+    assert wait_times == expected[2]
+    assert total_wait_time == expected[3]
+
+
+def test_event_building():
+    """Test the data relating to the timeline with an event still building."""
+    bug = BugzillaBug.get_or_create({
```
mprahl: Could you reuse `make_artifact` in this test?

estuary/models/errata.py
mprahl: Could you please remind me where this is needed in the frontend?

sarah256: For the creation time of the advisory, unless we would just used the created at property for it still, in which case it would no longer matter.

estuary/utils/story.py
sarah256: No, because if it goes into the if statement for the build, it will also later need to go into the else statement to get its completion time

estuary/utils/story.py
sarah256: Do you mean if it was stubbed out? Would a log.wanring be needed then?

estuary/utils/story.py
sarah256: Good idea.

mprahl:
@sarah256 let me know when this is ready for review again. It looks like some comments are still not addressed.

sarah256:
@mprahl This should *finally* be ready for review again lol...

estuary/api/v1.py
mprahl: @sarah256 where you able to try this?

estuary/models/errata.py
```diff
@@ -47,11 +47,21 @@ def display_name(self):
...
+    def attached_build_time(self, advisory, build):
+        """
+        Get the time that a build related to the advisory was attached.
+
+        :param node build: a Neo4j node representing an attached build
+        :return: the time the build was attached
+        :rtype: datetime object
+        """
+        rel = advisory.attached_builds.relationship(build)
+        return rel.time_attached
```
mprahl: What if `rel` is `None`? This happens when `build` is not attached to `advisory`.

estuary/utils/story.py
```diff
@@ -173,18 +175,24 @@ def get_wait_times(self, results):
...
             property_name = completion_times[artifact.__label__]
             completion_time = getattr(artifact, property_name)
-            if not completion_time or not next_artifact.timeline_timestamp:
+            if not completion_time or not next_artifact.timeline_datetime:
                 continue
 
-            next_artifact_start_time = datetime.strptime(next_artifact.timeline_timestamp,
-                                                         '%Y-%m-%dT%H:%M:%SZ')
+            if next_artifact.__label__.endswith('Advisory'):
+                if next_artifact.attached_build_time(next_artifact, artifact):
```
mprahl: If this evaluates to `False`, then `next_artifact_start_time` will be `None`. An exception will be raised when `next_artifact_start_time = next_artifact_start_time.replace(tzinfo=None)` is executed.

estuary/api/v1.py
```diff
@@ -227,12 +235,17 @@ def _get_partial_stories(results, reverse=False):
 
     # Adding the artifact itself if its story is not available
     if not all_results:
+        base_instance = estuary.utils.story.BaseStoryManager()
+        wait_times, total_wait_time = base_instance.get_wait_times([item])
```
mprahl: Doesn't `get_wait_times` require 2 or more artifacts in the story?

estuary/api/v1.py
```diff
@@ -227,12 +235,17 @@ def _get_partial_stories(results, reverse=False):
...
         rv = {'data': [item.serialized_all], 'meta': {}}
         rv['meta']['story_related_nodes_forward'] = [0]
         rv['meta']['story_related_nodes_backward'] = [0]
         rv['meta']['requested_node_index'] = 0
         rv['meta']['story_type'] = story_manager.__class__.__name__[:-12].lower()
-        rv['meta']['wait_times'] = [0]
+        rv['meta']['wait_times'] = wait_times
+        rv['meta']['total_wait_time'] = total_wait_time
+        rv['meta']['total_processing_time'] = base_instance.get_total_processing_time([item])
```
mprahl: `format_story_results` uses `try/excepts` when it calls these. It'd be nice to have that here too.

estuary/utils/story.py
mprahl: @sarah256 I forget where we landed on this. Could you please remind me?

estuary/utils/story.py
```diff
@@ -195,9 +203,155 @@ def get_wait_times(self, results):
...
+                    'While calculating the total processing time, a %s with ID %s was encountered '
+                    'without a creation time.',
+                    artifact.__label__, id_num
+                )
+                continue
+
+            if artifact.__label__.endswith('KojiBuild'):
+                build = artifact
+
+            if artifact.__label__.endswith('Advisory'):
```
mprahl: Can you make this `elif`?

estuary/utils/story.py
```diff
@@ -195,9 +203,155 @@ def get_wait_times(self, results):
...
+            processing_time = completion_time - creation_time
+
+            if processing_time.total_seconds() < 0:
+                id_num = getattr(artifact, artifact.unique_id_property + '_')
+                log.warning(
+                    'A negative processing time was calculated, with a %s with ID %s.',
+                    artifact.__label__, id_num
+                )
+
+            total += processing_time.total_seconds()
```
mprahl: If the `if processing_time.total_seconds():` is `True`, then should this get executed?

estuary/utils/story.py
```diff
@@ -195,9 +203,155 @@ def get_wait_times(self, results):
...
+
+        if total < 0:
+            id_num = getattr(artifact, artifact.unique_id_property + '_')
+            log.warning(
+                'A negative total processing time was calculated, in a story ending with a %s with '
+                'ID %s.',
+                artifact.__label__, id_num
+            )
+        return total
```
mprahl: If `if total < 0:` is `True`, should we just return 0? I don't know if the frontend would support negative seconds. This point is moot though depending on your decision on the comment:
https://github.com/release-engineering/estuary-api/pull/254/files#r306845653

tests/api/test_timeline.py
```diff
@@ -0,0 +1,320 @@
...
+import pytz
+import pytest
+
+import estuary.utils.story
+from tests import helpers
+
+
+def test_full_timeline():
+    """Test the data relating to the timeline with a full story."""
+    bug = helpers.make_artifact('BugzillaBug', **{
```
mprahl: Optional: You could just do:
```python
bug = helpers.make_artifact('BugzillaBug', id_='1111', creation_time=datetime(2019, 1, 1, 0, 0, 0))
```

tests/api/test_timeline.py
```diff
@@ -0,0 +1,320 @@
...
+        'creation_time': None
+    })
+
+    results = [bug, commit, build]
+    base_instance = estuary.utils.story.BaseStoryManager()
+    base_instance.get_total_lead_time(results)
+    base_instance.get_total_processing_time(results)
+
+    assert caplog.record_tuples == [
+        ('estuary', 30, 'While calculating the total lead time, a BugzillaBug with ID 1111 was '
```
mprahl: Instead of `30`, could you use `logging.WARNING`? It makes it easier to read.

tests/helpers.py
```diff
@@ -0,0 +1,148 @@
...
+            'state': None,
+            'version': None
+        }
+    else:
+        return
+
+    args.update(**kwargs)
+
+    artifact_class = names_to_model[label]
+    # import pdb;pdb.set_trace()
```
mprahl: Could you delete this line?

tests/helpers.py
```diff
@@ -0,0 +1,148 @@
...
+            'module_stream': '1.4',
+            'module_version': '20180805121332',
+            'name': '389-ds',
+            'release': '20180805121332.a2037af3',
+            'start_time': datetime(2019, 1, 1, 0, 0, 40),
+            'state': None,
+            'version': None
+        }
+    else:
+        return
```
mprahl: You should raise an exception here instead

estuary/utils/story.py
sarah256: What do you think we should put in the else clause? Should we just continue to the next iteration of the loop?

estuary/api/v1.py
sarah256: If there are less than two, it returns [0], 0

estuary/utils/story.py
sarah256: I realized that if we look at each individual processing time, and  check if each one is negative, we don't even need to check the total time because it cannot be negative, since all the negatives are skipped with each iteration.

estuary/utils/story.py
mprahl: That works for me.

estuary/utils/story.py
```diff
@@ -173,18 +175,26 @@ def get_wait_times(self, results):
...
                 continue
 
-            next_artifact_start_time = datetime.strptime(next_artifact.timeline_timestamp,
-                                                         '%Y-%m-%dT%H:%M:%SZ')
+            if next_artifact.__label__.endswith('Advisory'):
+                if next_artifact.attached_build_time(next_artifact, artifact):
+                    next_artifact_start_time = next_artifact.attached_build_time(next_artifact,
+                                                                                 artifact)
+                else:
+                    continue
```
mprahl: Could you add a log statement here explaining why it was skipped?

estuary/utils/story.py
```diff
@@ -195,9 +205,157 @@ def get_wait_times(self, results):
...
+    def get_total_processing_time(self, results):
+        """
+        Get the total time spent processing the story.
+
+        :param list results: contains inflated results from Neo4j
+        :return: the seconds of total time spent processing
+        :rtype: int
+        """
+        total = 0
+        build = None
```
mprahl: Could you add a comment as to what this variable is for please? It just makes the code easier to read.

estuary/utils/story.py
```diff
@@ -195,9 +205,157 @@ def get_wait_times(self, results):
...
+            end_time = getattr(last_artifact, start_time_key)
+
+        # Remove timezone info so that both are offset naive and thus able to be subtracted
+        start_time = start_time.replace(tzinfo=None)
+        end_time = end_time.replace(tzinfo=None)
+        total = end_time - start_time
+        if total.total_seconds() < 0:
+            id_num = getattr(last_artifact, last_artifact.unique_id_property + '_')
+            log.warning(
+                'A negative total lead time was calculated, in a story ending with a %s with '
```
mprahl: Could you reword the log message so that it mentions the first and last artifact and their IDs? This then gives us enough context to debug.

estuary/api/v1.py
```diff
@@ -115,12 +114,22 @@ def _get_partial_story(results, reverse=False):
...
         rv['meta']['requested_node_index'] = 0
         rv['meta']['story_type'] = story_manager.__class__.__name__[:-12].lower()
-        rv['meta']['wait_times'] = [0]
+        rv['meta']['wait_times'] = wait_times
+        rv['meta']['total_wait_time'] = total_wait_time
+        rv['meta']['total_processing_time'] = None
+        rv['meta']['total_lead_time'] = None
+        try:
+            rv['meta']['total_processing_time'] = base_instance.get_total_processing_time([item])
+            rv['meta']['total_lead_time'] = base_instance.get_total_lead_time([item])
```
mprahl: I just realized that the total lead time will always be 0 here since it's just a single artifact. I think you can just set it directly and move it outside of the try and except clause.

estuary/api/v1.py
```diff
@@ -227,12 +236,22 @@ def _get_partial_stories(results, reverse=False):
...
```
mprahl: Same comment as above

estuary/utils/story.py
sarah256: Oh yeah I talked to Ralph. He said "I'm okay with the current behavior. of skipping the stubbed artifact and having the total time be slightly off"

mprahl:
@sarah256 could you please fix the tests?

sarah256:
@mprahl This should be good for review again

PR #255, opened 2019-07-15
Use datetime objects instead of strings when creating Freshmaker events

mprahl:


mprahl:
> Looks good! But what is the benefit of doing this?

Neomodel expects a datetime object and not a string, so the scrapers were failing with a traceback.

PR #258, opened 2019-07-16
Replace existing attached relationships when time_attached is missing or different

mprahl:
If replace isn't used, a second relationship is created to the node.

### DIFF
--- a/scrapers/errata.py
+++ b/scrapers/errata.py
@@ -113,13 +113,13 @@ class ErrataScraper(BaseScraper):
                     nvr = '-'.join((associated_build['name'], associated_build['version'],
                                     associated_build['release']))
 
-                    sql = r"""\
+                    sql = """\
                         SELECT created_at
                         FROM Errata_public.errata_brew_mappings
                         INNER JOIN Errata_public.brew_builds
                         ON Errata_public.errata_brew_mappings.brew_build_id
                             = Errata_public.brew_builds.id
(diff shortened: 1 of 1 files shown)


mprahl:
@sarah256 once this is merged, could you please make the same fix in estuary-updater?

scrapers/errata.py
```diff
@@ -113,19 +113,28 @@ def update_neo4j(self, advisories):
...
+                        WHERE errata_id = {0} AND current = 1 AND nvr = '{1}';
                     """.format(advisory['id'], nvr)
                     log.info('Getting the time build {0} was added to advisory {1}'.format(
                         nvr, advisory['id']))
-                    time_attached = self.teiid.query(sql)
-
-                    adv.attached_builds.connect(build, {'time_attached': time_attached})
+                    time_attached_result = self.teiid.query(sql)
+                    if time_attached_result:
+                        time_attached = time_attached_result[0].get('created_at')
```
sarah256: Wouldn't the SQL query already grab just the 'created_at'?

sarah256:
Just one little comment, but looks good elsewhere!

scrapers/errata.py
mprahl: It returns a list with a single entry. You can see that in the documentation:
https://estuary-api.readthedocs.io/en/latest/module_documentation/scrapers.html#scrapers.teiid.Teiid.query

PR #260, opened 2019-07-18
On stubbed out artifacts, tell user that timestamp information is unavailable

sarah256:


sarah256:
Closed because it was decided to make this change on the front end instead.

PR #262, opened 2019-07-24
Sort Freshmaker events by timestamps instead of by IDs in recents endpoint

sarah256:


mprahl:
@sarah256 this looks good but you'll need to adjust the frontend since the sorting occurs there in the table.

sarah256:
@mprahl Could you clarify what part of the front end you mean? Running it with the back end and just this change, it seems to be showing the proper thing, but I may be mistaken.

mprahl:
@sarah256 that's what I was referring to. I got confused as to the purpose of this PR. I'll review it again.

estuary/utils/recents.py
```diff
@@ -22,7 +22,7 @@ def get_recent_nodes():
     :rtype: tuple
     """
     timestamp_dict = {
-        FreshmakerEvent.__label__: 'id',
+        FreshmakerEvent.__label__: 'time_done',
```
mprahl: I think it's fine to use `time_done`, but this would limit it to just `FreshmakerEvent` nodes that have completed. Is that okay?

sarah256: @mprahl I thought about that too. I wonder if using the time it was created would be better?

mprahl: @sarah256 that's a good idea. We can change it later if needed.

PR #264, opened 2019-08-01
Add a flag for when processing time cannot be fully calculated due to lack of data

sarah256:


estuary/utils/story.py
```diff
@@ -225,9 +225,10 @@ def get_total_processing_time(self, results):
         Get the total time spent processing the story.
 
         :param list results: contains inflated results from Neo4j
-        :return: the seconds of total time spent processing
-        :rtype: int
+        :return: the seconds of total time spent processing with a flag for inaccurate calculations
+        :rtype: tuple
         """
+        flag = 0
```
mprahl: Could you use a boolean instead?

estuary/utils/story.py
```diff
@@ -252,6 +253,7 @@ def get_total_processing_time(self, results):
                     'without a creation time.',
                     artifact.__label__, id_num
                 )
+                flag = 1
```
mprahl: Can you add a unit test for this case?

sarah256:
@mprahl This should be ready for review again

PR #265, opened 2019-08-12
Skip advisories with no creation time when calculating processing time

sarah256:
Sometimes advisories do not have time data for their associated build(s). We need to be able to catch when this happens so that the processing time can still be calculated, instead of just producing a traceback.


yashvardhannanavati:
LGTM!

PR #266, opened 2019-08-13
Add a flag to indicate processing time calculation may be incorrect if an advisory is not associated with a build

sarah256:


gnaponie:
I don't quite understand how that is going to being shown. I mean what's the impact of the change. Can you explain it if possible?

sarah256:
> I don't quite understand how that is going to being shown. I mean what's the impact of the change. Can you explain it if possible?

There is a corresponding front end change that I am still working on. Should've mentioned it, my bad! Basically, the flag will indicate if a toast notification warning should pop up or not in the front end that warns users that the calculation may be incorrect.

gnaponie:
Oh I see! This looks good then IMO.

PR #268, opened 2019-09-13
Allow an LDAP group to be used as a user whitelist for authorization

mprahl:
This addresses FACTORY-4860

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,3 +88,32 @@ Get a resource from Neo4j.
 
 ## Code Documentation
 To document new files, please check [here](https://github.com/release-engineering/estuary-api/tree/master/docs).
+
+
+## Authorization
+
+If authentication is enabled, Estuary can authorize users based on their employee type and a user
+whitelist configured through the membership of an LDAP group.
+
+### Employee Type
+
+You may set the list of valid employee types with the configuration item `EMPLOYEE_TYPES`. These
+employee types map to the `employeeType` LDAP attribute of the user that is added to the OpenID
+Connect token received by Estuary.
+
+### Configuring the Whitelist
+
+To configure a whitelist of users, they must be part of an LDAP group configured with Estuary. The
+following configuration items are required:
+
+* `LDAP_URI` - the URI to the LDAP server to connect to in the format of
+    `ldaps://server.domain.local`.
+* `LDAP_EXCEPTIONS_GROUP_DN` - the distinguished name to the LDAP group acting as the whitelist.
+
+The following configuration items are optional:
+
+* `LDAP_CA_CERTIFICATE` - the path to the CA certificate that signed the certificate used by the
+    LDAP server. This only applies if you are using LDAPS. This defaults to
+    `/etc/pki/tls/certs/ca-bundle.crt`.
+* `LDAP_GROUP_MEMBERSHIP_ATTRIBUTE` - the LDAP attribute that represents a user in the group. This
+    defaults to `uniqueMember`.
--- a/estuary/authorization.py
+++ b/estuary/authorization.py
@@ -1,0 +1,18 @@
+# SPDX-License-Identifier: GPL-3.0+
+
+from __future__ import unicode_literals
+
+from flask import current_app
+
+
+def is_user_authorized(employee_type):
+    """
+    Verify the user is authorized to access the application.
+
+    :param str employee_type: the employee type from the user's token
+    :return: a boolean that determines if the user is authorized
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -88,3 +88,32 @@ Get a resource from Neo4j.
...
+    `ldaps://server.domain.local`.
+* `LDAP_EXCEPTIONS_GROUP_DN` - the distinguished name to the LDAP group acting as the whitelist.
+
+The following configuration items are optional:
+
+* `LDAP_CA_CERTIFICATE` - the path to the CA certificate that signed the certificate used by the
+    LDAP server. This only applies if you are using LDAPS. This defaults to
+    `/etc/pki/tls/certs/ca-bundle.crt`.
+* `LDAP_GROUP_MEMBERSHIP_ATTRIBUTE` - the LDAP attribute that represents a user in the group. This
+    defaults to `uniqueMember`.
```
sarah256: What is `unigueMember`? Is it the current member or something else?

README.md
```diff
@@ -88,3 +88,32 @@ Get a resource from Neo4j.
...
+Connect token received by Estuary.
+
+### Configuring the Whitelist
+
+To configure a whitelist of users, they must be part of an LDAP group configured with Estuary. The
+following configuration items are required:
+
+* `LDAP_URI` - the URI to the LDAP server to connect to in the format of
+    `ldaps://server.domain.local`.
+* `LDAP_EXCEPTIONS_GROUP_DN` - the distinguished name to the LDAP group acting as the whitelist.
```
sarah256: What do you mean by 'distinguished name'? Is this something users will generally understand what you are referring to, since it could just be me not knowing.

README.md
```diff
@@ -88,3 +88,32 @@ Get a resource from Neo4j.
...
 ## Code Documentation
 To document new files, please check [here](https://github.com/release-engineering/estuary-api/tree/master/docs).
+
+
+## Authorization
+
+If authentication is enabled, Estuary can authorize users based on their employee type and a user
+whitelist configured through the membership of an LDAP group.
+
+### Employee Type
```
sarah256: Are there too many different types to be listed here?

estuary/authorization.py
```diff
@@ -0,0 +1,92 @@
...
+
+    If the LDAP search fails, an empty set is returned.
+
+    :return: a set of usernames
+    :rtype: set
+    :raise InternalServerError: if a required configuration value is not set or the connection to
+        the LDAP server fails
+    """
+    # Import this here so it's not required for deployments with auth disabled
+    import ldap3
```
sarah256: Why can't this be put at the top of the file regardless of whether auth is disabled or not? If auth was disabled, wouldn't Estuary not use this file anyway?

estuary/authorization.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def _get_exception_users():
+    """
+    Get the list of users that are explicitly whitelisted.
+
+    If the LDAP search fails, an empty set is returned.
+
+    :return: a set of usernames
+    :rtype: set
```
sarah256: What is the advantage of using a set here?

README.md
mprahl: It's just the LDAP attribute that Rover groups use to display the members of the group. It'd be just any member in that group.

README.md
mprahl: It's a string that uniquely identifies an entry in the LDAP directory. It's an implementation detail that the user won't be aware of.

README.md
mprahl: I don't want to list the employee types since it's specific to Red Hat's private infrastructure.

estuary/authorization.py
mprahl: This function is used by `estuary.utils.general` regardless of if authentication is enabled or not. I'll import it in `login_required` instead and move this to the top.

mprahl: Actually, this way it allows you to enable authentication and not require `ldap3` to be installed if the configuration `LDAP_EXCEPTIONS_GROUP_DN` is not set. I'll leave it as is.

estuary/authorization.py
mprahl: No advantage other than it let's the caller of the function know that there are no duplicate values.

README.md
sarah256: Oops! By users I mean like people using the whitelist config stuff. The only reason I ask is because it might be confusing to people reading it, and thus beneficial to specify what you mean here.

PR #271, opened 2019-09-23
Add 'operator' property to ContainerKojiBuild model to determine whether it is an operator

sarah256:
@lslebodn


gnaponie:
It looks good! But could you add also a test with "True" and one with "False" as values and not just "None"?

scrapers/koji.py
```diff
@@ -198,6 +198,11 @@ def update_neo4j(self, builds):
                 extra_json = {}
 
             if self.is_container_build(build_dict):
+                if extra_json.get('typeinfo').get('operator_manifests').get('archive'):
```
lslebodn: I assume that line throw an exception for images which are not operator
```
>>> extra_json.get('container_koji_task_id')
23678038
>>> extra_json.get('typeinfo').get('operator_manifests').get('archive')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'get'
>>> 
>>> extra_json.get('typeinfo')
>>> extra_json.get('typeinfo') is None
True
```

Following part should work better. But would be good o add unit tests
```
try:
    extra_json.get('typeinfo').get('operator_manifests').get('archive')
    has_operator = True
except AttributeError:
    has_operator = Fase
```

scrapers/koji.py
```diff
@@ -198,6 +198,12 @@ def update_neo4j(self, builds):
                 extra_json = {}
 
             if self.is_container_build(build_dict):
+                try:
+                    extra_json.get('typeinfo').get('operator_manifests').get('archive')
```
mprahl: My preference is to do the following:
```python
build_params['operator'] = bool(extra_json.get('typeinfo', {}).get('operator_manifests', {}).get('archive'))
```

Then you don't need the try/except.

PR #276, opened 2019-12-30
Modify Estuary scraper to use v2 API

sarah256:


mprahl:
@sarah256 there's part of the code that has `# TODO: Remove this once FACTORY-3955 is resolved`. Could you remove that code since this PR addresses that issue?

scrapers/freshmaker.py
```diff
@@ -87,7 +87,10 @@ def query_api_and_update_neo4j(self):
 
                 event.conditional_connect(event.triggered_by_advisory, advisory)
 
-                for build_dict in fm_event['builds']:
+                event_builds_url = ("https://freshmaker.engineering.redhat.com/api/2/builds/"
```
mprahl: Could you use single quotes to match the rest of the codebase?

scrapers/freshmaker.py
```diff
@@ -18,7 +18,7 @@
 class FreshmakerScraper(BaseScraper):
     """Scrapes the Freshmaker API."""
 
-    freshmaker_url = "https://freshmaker.engineering.redhat.com/api/1/events/?per_page=50"
+    freshmaker_url = "https://freshmaker.engineering.redhat.com/api/2/events/?per_page=50"
```
mprahl: Optional: Could you use single quotes to match the rest of the codebase? I know you didn't introduce this, but it'd be nice to fix this since you're already modifying this.

scrapers/freshmaker.py
```diff
@@ -87,7 +78,10 @@ def query_api_and_update_neo4j(self):
 
                 event.conditional_connect(event.triggered_by_advisory, advisory)
 
-                for build_dict in fm_event['builds']:
+                event_builds_url = ('https://freshmaker.engineering.redhat.com/api/2/builds/'
```
mprahl: @sarah256 it seems like this API endpoint uses pagination. Could you make sure your code gets all the builds for every page? It might also help to increase the `per_page` limit so that there is less network IO involved.

scrapers/freshmaker.py
```diff
@@ -87,7 +78,15 @@ def query_api_and_update_neo4j(self):
...
```
mprahl: How about the following approach? It's a little awkward since Python doesn't have a do while loop though.
```python
event_builds_url = 'https://freshmaker.engineering.redhat.com/api/2/builds/?event_id={0}&per_page=50'.format(fm_event['id'])
builds = []
while True:
    rv = session.get(event_builds_url, timeout=60).json()
    builds.extend(rv['items'])
    if rv['meta']['next']:
        event_builds_url = rv['meta']['next']
    else:
        break
```

There's nothing technically wrong with your approach, but I find it unconventional that you are altering the JSON returned by the Freshmaker API. In addition, the duplicate the `session.get` method call with your approach is unnecessary in this case. Feel free to change the above example to your style though (e.g. changing variable names).

PR #278, opened 2020-06-23
Add mypy for testing and CI

mulaievaRH:
Add the type checker in Travis CI

JIRA: https://projects.engineering.redhat.com/browse/CLOUDWF-771

### DIFF
--- a/docker/Dockerfile-tests
+++ b/docker/Dockerfile-tests
@@ -12,6 +12,7 @@ RUN dnf -y install \
   python3-flask-oidc \
   python3-ldap3 \
   python3-mock \
+  python3-mypy \
   python3-neomodel \
   python3-prometheus_client \
   python3-pytest \
@@ -22,4 +23,4 @@ RUN dnf -y install \
 VOLUME /src
 WORKDIR /src
 # Inspired from https://github.com/neo4j-contrib/neomodel/blob/master/tests-with-docker-compose.sh
-CMD ["sh", "-c", "while ! nc -z -w 2 neo4j 7687; do sleep 1; done; pytest-3 -vvv --cov-report term-missing --cov=estuary tests/ && flake8"]
+CMD ["sh", "-c", "while ! nc -z -w 2 neo4j 7687; do sleep 1; done; pytest-3 -vvv --cov-report term-missing --cov=estuary tests/ && flake8 && mypy --ignore-missing-imports ."]
--- a/estuary/config.py
+++ b/estuary/config.py
@@ -2,6 +2,8 @@
 
 from __future__ import unicode_literals
 
+from typing import List, Optional
+
 
 class Config(object):
     """The base Estuary application configuration."""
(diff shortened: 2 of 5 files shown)


gnaponie:
+1
@mprahl could you also review this?

estuary/config.py
```diff
@@ -22,10 +23,10 @@ class Config(object):
     CORS_ORIGINS = ['http://localhost:4200']
     STORY_MANAGER_SEQUENCE = ['ModuleStoryManager', 'ContainerStoryManager']
     ENABLE_AUTH = False
-    OIDC_INTROSPECT_URL = None
-    OIDC_CLIENT_ID = None
-    OIDC_CLIENT_SECRET = None
-    EMPLOYEE_TYPES = []
+    OIDC_INTROSPECT_URL = None  # type: str
```
mprahl, line 26: Could we use the Python 3 syntax please?

https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#built-in-types

.travis.yml
```diff
@@ -1,5 +1,22 @@
-language: generic
-sudo: required
-services: docker
-script:
-  - bash scripts/run-tests.sh
+matrix:
+  include:
+    - language: python
```
mprahl, line 3: Although this is quite nice to speed up TravisCI runs, the documentation tells users to run `scripts/run-tests.sh` and it will not run the mypy test in that case.

You can run mypy in here instead so that it runs both in TravisCI and locally:
https://github.com/release-engineering/estuary-api/blob/9b090856996646294043765b669dd09edb5c0ff4/docker/Dockerfile-tests#L25

If you feel inspired, you could use a matrix and overwrite the run command in the command so that one matrix is for flake8, the other for pytest, and the last one for mypy.

mprahl:
Could you please add a section to the "Code Styling" section in the read me mentioning that mypy is now used?

mulaievaRH:
@mprahl , PTAL one more time ))

estuary/config.py
```diff
@@ -1,7 +1,8 @@
 # SPDX-License-Identifier: GPL-3.0+
-
```
mprahl, line 2: Could you please add this new line back that you accidentally removed?

setup.py
```diff
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0+
 from setuptools import setup, find_packages
 
-requirements = []
```
mprahl, line 4: This change is fine but it's not necessary and unrelated to the PR. Could you please readd this line?

docker/Dockerfile-tests
```diff
@@ -17,9 +17,10 @@ RUN dnf -y install \
   python3-pytest \
   python3-pytest-cov \
   python3-flake8-docstrings \
+  python3-mypy \
```
mprahl, line 20: Could you please alphabetize this?

setup.py
mulaievaRH: I had a mypy error due this line. https://travis-ci.org/github/release-engineering/estuary-api/builds/702283638
So I should delete this line or add a type definition here. So I've decided to delete  it.

PR #280, opened 2021-03-12
Spin local dev environment with docker-compose

lioramilbaum:
Upgrade: neo4j, apoc, api parent image (fedora 32)

### DIFF
--- a/.github/workflows/functional.yml
+++ b/.github/workflows/functional.yml
@@ -1,0 +1,18 @@
+name: Functional
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.9
+      - name: Functional
+          make test
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+.PHONY: up build down dependencies test logs
+
+down:
+	docker-compose down -v
+
+build:
+	docker-compose build
+
+up: build
+	docker-compose up -d
+
+logs: up
+	docker-compose logs -f
+
+dependencies:
+	pip install --upgrade pip
+	pip install -r tests/requirements.txt
+
+test: dependencies up
+	pytest --noconftest tests/functional
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,42 @@
+version: '3.3'
+services:
+  api:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile-api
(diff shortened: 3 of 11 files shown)


lioramilbaum:
@mzidek-gh @hanzz take a look

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+    expose:
+      - 8080
+    environment:
+      NEO4J_URI: bolt://neo4j:neo4j@db:7687
+      DEV: "true"
+    depends_on:
+      db:
+        condition: service_healthy
+  db:
+    image: neo4j:3.4-enterprise
```
mprahl, line 19: Will 3.5 work?

Makefile
```diff
@@ -0,0 +1,17 @@
...
+	docker-compose build
+
+up: build
+	docker-compose up -d
+
+dependencies:
+	pip install --upgrade pip
+	pip install -r tests/requirements.txt
+
+test: dependencies up
```
mprahl, line 16: Could we wrap this in tox so that dependencies are in a virtualenv and functional test and we can then call our additional tests such as linting and unit in the same way?

Makefile
```diff
@@ -0,0 +1,17 @@
+.PHONY: up build down dependencies test
+
+down:
+	docker-compose down
+
+build:
+	docker-compose build
+
+up: build
+	docker-compose up -d
```
mprahl, line 10: I prefer it without the `-d` flag so you can see the logs. Any chance we can make that flag optional?

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: '3.3'
```
mprahl, line 1: Could you add a volume for Neo4j's database files?

.github/workflows/functional.yml
```diff
@@ -0,0 +1,18 @@
+name: Functional
```
mprahl, line 1: Very cool. Let me know what you need from me to enable this.

lioramilbaum: GitHub Actions will start executing the workflow on future PRs, as soon as this one is merged.

Makefile
lioramilbaum: We can have a different target for logs. Does that make sense?

Makefile
lioramilbaum: Sure. Lets do that in the next PR. I don't want it to block this PR from been merged.

docker-compose.yml
lioramilbaum: I think I've checked it and it didn't.  I think I've filed an issue for upgrading neo4j.

Makefile
mprahl: Sure

mprahl (approved):

PR #287, opened 2021-03-19
Bump neo4j release

lioramilbaum:


docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       db:
         condition: service_healthy
   db:
-    image: neo4j:3.4-enterprise
+    image: neo4j:enterprise
```
mprahl, line 19: Can we stay on the latest 3.x for now please? This matches the current prod environment.

lioramilbaum: Sure. Although, I do want to check if the newer release has some new capabilities which can be useful.
BTW, where is the code which deploys the db tracked?

lioramilbaum:
@mprahl Can you think of a test which can validate that a community version is working properly?

mzidek-gh:
Just one nitpick from my side. Please put the changes in the  .gitpod.yml into a separate commit as they are not related to the version bump.

docker-compose.yml
lioramilbaum: Done

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       db:
         condition: service_healthy
   db:
-    image: neo4j:3.4-enterprise
+    image: neo4j:3.5.26
```
mprahl, line 19: Were you able to test a query by any chance? I'm just worried that the [APOC plugin](https://github.com/neo4j-contrib/neo4j-apoc-procedures) also needs to be updated.

mprahl, line 19: Is there a tag that would allow us to use any patch version of 3.5?

lioramilbaum: There is. Why do you think a floating tag is better than a pinned version?

mprahl: This would allow us to not have to update the tag for every patch version release of Neo4j and patch versions should always be safe to upgrade to, especially in a dev environment.

lioramilbaum: I did not. We can implement a functional test.

The APOC plugin is picked up by neo4j automatically.  
https://neo4j.com/labs/apoc/4.0/installation/#docker

lioramilbaum: The neo4j release should be the same in all environments, unless we are considering upgrading prod environment. That is the reason why I want to take a look at the neo4j deployment code.
With that, there is no reason to update the tag if we do not plan to upgrade the prod.
On the other hand, a pinned version helps spin a reproducible environment for troubleshooting. This is similar to how we pin PyPi packages in requirements.txt files.

mprahl: @lioramilbaum Could you please confirm that is the case for 3.x too?

mprahl: Okay, no problem. We can keep it pinned.

lioramilbaum:
```Status: Downloaded newer image for neo4j:3.5.26
Creating db ... done
Creating api ... done
(estuary) [lmilbaum@lmilbaum estuary-api]$ docker logs db
Fetching versions.json for Plugin 'apoc' from https://neo4j-contrib.github.io/neo4j-apoc-procedures/versions.json
Installing Plugin 'apoc' from https://github.com/neo4j-contrib/neo4j-apoc-procedures/releases/download/3.5.0.15/apoc-3.5.0.15-all.jar to /var/lib/neo4j/plugins/apoc.jar 
Applying default values for plugin apoc to neo4j.conf
Skipping dbms.security.procedures.unrestricted for plugin apoc because it is already set
Active database: graph.db
Directories in use:
  home:         /var/lib/neo4j
  config:       /var/lib/neo4j/conf
  logs:         /logs
  plugins:      /var/lib/neo4j/plugins
  import:       /var/lib/neo4j/import
  data:         /var/lib/neo4j/data
  certificates: /var/lib/neo4j/certificates
  run:          /var/lib/neo4j/run
Starting Neo4j.
2021-03-23 06:46:22.392+0000 INFO  ======== Neo4j 3.5.26 ========
2021-03-23 06:46:22.404+0000 INFO  Starting...
2021-03-23 06:46:29.807+0000 INFO  Bolt enabled on 0.0.0.0:7687.
2021-03-23 06:46:31.117+0000 INFO  Started.
2021-03-23 06:46:32.029+0000 INFO  Remote interface available at http://localhost:7474/```

lioramilbaum: Added another workflow - infra, for testing infra issues. Also added a test which validates the plugin is installed.

mprahl:
@lioramilbaum this looks good. Could you please rebase?

PR #290, opened 2021-03-21
tox

lioramilbaum:


mzidek-gh:
Hi, please make the commit message at least a sentence so it is clear what was done without reading the code :)

tests/requirements.txt
```diff
@@ -1,2 +1,6 @@
 docker-compose==1.28.5
+flask-oidc==1.4.0
```
mprahl, line 2: How did you determine this subset of dependencies?

lioramilbaum: I invoked tests multiple and added the packages one by one.

mprahl: Shouldn't this just also include the requirements from `requirements.txt`?

lioramilbaum: Maybe,  I didn't spend to much thoughts on that. We might want to switch to pipenv or poetry.

mprahl: Could you please only include the dependencies not in `requirements.txt`?

lioramilbaum: Done

lioramilbaum:
I am wondering why the travis CI is not running. @mprahl can you please check?

mprahl:
@lioramilbaum we ran out of free credits, so we need to migrate to GitHub actions. I just recently had to do this for another project in the namespace.

tests/requirements.txt
```diff
@@ -1,2 +1,4 @@
 docker-compose==1.28.5
+mock==4.0.3
```
mprahl, line 2: We really should just change the imports to use `unittest.mock` since we don't need to support Python 2.7 anymore, but this is fine for now.

lioramilbaum:
> @lioramilbaum we ran out of free credits, so we need to migrate to GitHub actions. I just recently had to do this for another project in the namespace.

OK. I will migrate the code to GitHub Actions in in following PRs.

Makefile
```diff
@@ -12,9 +12,6 @@ up: build
...
 	docker-compose logs -f
 
-dependencies:
-	pip install --upgrade pip
-	pip install -r tests/requirements.txt
-
-test: dependencies up
-	pytest --noconftest tests/functional
+test: up
+	pip install tox==3.23.0
```
mprahl, line 16: I think most people running this locally would have installed tox via an RPM.

How about not pinning the version exactly so that the one installed via an RPM works and install it in the user directory?
```suggestion
	pip install --user tox>3.0.0
```

lioramilbaum: I am not really following the RPM issue. Aren't we dropping that in favor of PyPi packages?
Pinning package versions enabled reproducible environment. Even test environment should be reproducible.
I have seen examples of new versions which didn't provide backwards compatibility which broke CI.

mprahl: @lioramilbaum yes, but `tox` creates the virtual environment which is the one package that we should allow not from PyPi since many people do not like installing system dependencies via PyPi.

lioramilbaum: This is a valid point. GitHub Actions spins an ubuntu VM.  Take a look at the instructions:
https://askubuntu.com/questions/423716/how-do-i-install-tox

mprahl: @lioramilbaum yes, but developers will want to run `make test` locally too. We should accommodate for both situations.

hanzz: @lioramilbaum , I think @mprahl 's point is that with your change, we can actually let developers to run "make test" locally without influencing their system which is good thing and improvement from previous state :). With the previous version (without your PR), the `pip install -r` caused system versions (coming from RPMs) of all those required python modules to be overridden by the versions defined in the `requirements.txt`. So if you executed `make test`, it would install those modules and binaries to `~/.local/bin/` (for example) and use that instead of the version in /usr/bin everytime you try to execute them on your user account.

For example, I have `python3-pytest-6.0.2-1.fc33.noarch` installed on my system, but after running the old `make test`, I end up with following version everytime I execute pytest (no matter if I want to work on estuary or other project):

```
$ pytest --version
pytest 6.2.2`
```

The benefit of using tox is that it creates virtual environment for your test so you do not mess up your user account with custom versions of python modules like that. And that's why I think @mprahl wants you to not pinpoint particular tox version (and maybe also docker-compose), because if you pinpoint it, it will install this different version and it will be used everytime you run "tox" using your user account instead of the version coming from RPM. If you don't pinpoint the version and just user `pip install tox>3.0.0` with `python3-tox` RPM installed, it will not install anything and just use the system version of tox coming from RPM, because it satisfies this version required in pip install.

I think for tests we should care about the python modules *in* the virtual env created by tox, but the version of tox or docker-compose is not that important to be worth changing the default tox or docker-compose version for developer running `make test`.

hanzz:
Looks good to me now, @mprahl , what do you think? :)

mprahl (approved):

mprahl:
Looks good. We should also eventually migrate to using tox for the unit tests, though since it requires docker-compose (often requires root), this may not be a clean solution.

PR #297, opened 2021-04-05
Sort imports with isort

lioramilbaum:


mprahl:
@lioramilbaum this is a cool tool. One of the files seems to have some style changes done to it as well. Could you please exclude those?

FYI, I'm also fine with using `black` to format everything.

.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
+known_first_party = estuary
```
mprahl, line 2: Should this also include `scrapers` and `tests`?

mprahl: Why did you add estuary there then if it can auto detect?

lioramilbaum: because it didn't auto detect

PR #298, opened 2021-04-05
ProductVersion model

lioramilbaum:
Product Version is the primary filter for container automation data
/cc @mzidek-gh

### DIFF
--- a/estuary/models/productversion.py
+++ b/estuary/models/productversion.py
@@ -1,0 +1,9 @@
+from neomodel import StringProperty
+
+from estuary.models.base import EstuaryStructuredNode
+
+
+class ProductVersion(EstuaryStructuredNode):
+    """Definition of Product Version in Neo4j."""
+
+    name = StringProperty(unique_index=True)
--- a/tests/functional/test_ProductVersion.py
+++ b/tests/functional/test_ProductVersion.py
@@ -1,0 +1,26 @@
+import json
+
+import pytest
+import requests
+
+from estuary.models.productversion import ProductVersion
+
+
+@pytest.mark.parametrize("name", [("8.4.0.0"), ("8.5.0.0")])
+def test_load_ProductVersion(name):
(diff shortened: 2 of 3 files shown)


estuary/models/productversion.py
```diff
@@ -0,0 +1,9 @@
+from neomodel import StringProperty
+
+from estuary.models.base import EstuaryStructuredNode
+
+
+class ProductVersion(EstuaryStructuredNode):
+    """Definition of RHEL Product Version in Neo4j."""
+
+    name = StringProperty()
```
mprahl, line 9:
```suggestion
    name = StringProperty(unique=True, index=True)
```

tests/functional/test_ProductVersion.py
```diff
@@ -0,0 +1,25 @@
+import json
+
+import pytest
+import requests
+from estuary.models.productversion import ProductVersion
```
mprahl, line 5: I think isort (#297) would change this.

lioramilbaum: fixed

estuary/models/productversion.py
lioramilbaum: According to neomodel documentation, unique_index argument is the same as using unique and index arguments. WDYT?

mprahl: Sounds good!

mprahl (approved):

estuary/models/productversion.py
```diff
@@ -0,0 +1,9 @@
+from neomodel import StringProperty
+
+from estuary.models.base import EstuaryStructuredNode
+
+
+class ProductVersion(EstuaryStructuredNode):
+    """Definition of RHEL Product Version in Neo4j."""
```
mprahl, line 7: Sorry, I missed this before. I think it'd be best if this was just a generic concept of a product version.
```suggestion
    """Definition of Product Version in Neo4j."""
```

PR #300, opened 2021-04-13
Upgrade the base images to use UBI 8

mprahl:
Fedora 30 is not supported anymore and now
that dependencies are installed using PyPi,
the UBI 8 image can be used.

The first commit uses PyPi for the scrapers and tests dependencies.


docker/Dockerfile-api
```diff
@@ -1,4 +1,4 @@
-FROM fedora:30
+FROM registry.access.redhat.com/ubi8/ubi
```
lioramilbaum, line 1: Please pin the parent image tag

docker/Dockerfile-scrapers
```diff
@@ -1,4 +1,4 @@
-FROM fedora:30
+FROM registry.access.redhat.com/ubi8/ubi
```
lioramilbaum, line 1: same comment as before

docker/Dockerfile-tests
```diff
@@ -1,25 +1,18 @@
-FROM fedora:30
+FROM registry.access.redhat.com/ubi8/ubi
```
lioramilbaum, line 1: same comment.
BTW, have you considered using ubi-minimal image?

lioramilbaum (changes requested):
This PR is huge. It contains multiple efforts which are unrelated to the PR title. Please break the PR into smaller PRs. Each one for its own specific effort.

docker/Dockerfile-api
mprahl: @lioramilbaum I'd rather not pin it since Red Hat will ensure compatibility and thus I'd rather just always be on the latest UBI8 to inherit security fixes on deployments.

docker/Dockerfile-tests
mprahl: I'll give it a shot.

mprahl:
@lioramilbaum thanks for taking a look. I'll break up the PR.

docker/Dockerfile-api
lioramilbaum: Unpinned image tag will prevent us from reproducing the exact environment running in production (for example). Using ephemeral tags is an anti pattern as it comes to reproducible environments.

PR #304, opened 2021-04-19
Upgrade the base images to use UBI 8

mprahl:
Fedora 30 is not supported anymore and now
that dependencies are installed using PyPi,
the UBI 8 image can be used.


apaplaus (approved):

lioramilbaum (commented):
Have we agreed on using ephemeral tag?

mprahl:
@lioramilbaum yes, because I want to automatically pick up CVE fixes from the base image rather than have to update the tag every time. This is quite low risk since it is UBI 8 and the image only installs a few RPMs.

PR #310, opened 2021-06-06
Default to requiring SSL when connecting to Teiid

mprahl:


lmilbaum (changes requested):
Can you please provide (at least) unit test?

scrapers/teiid.py
```diff
@@ -31,14 +31,27 @@ def __init__(self, host, port, username, password):
...
+        sslrootcert='/etc/pki/tls/certs/ca-bundle.crt',
+    ):
         """
         Return an existing psycopg2 connection and establish it if needed.
 
         :param str db_name: the database name to get a connection to
         :kwarg bool force_new: forces a new database connection even if one already exists
         :kwarg int retry: the number of times to retry a failed connection. If this is not set,
             then the Teiid connection attempt will be repeated until it is successful.
+        :kwarg str sslmode: the Postgres SSL mode to use when connecting to Teiid. This defaults
```
apaplaus, line 49: Maybe it's worth mentioning what other options are acceptable for this parameter?

mulaievaRH (approved):

mprahl:
> Can you please provide (at least) unit test?

Unfortunately, the scrapers don't have unit tests since they were only written to be used temporarily until Estuary-Updater was written shortly after. The scrapers have stayed operational to account for message bus outages and/or Estuary-Updater bugs, but unit tests were never added retroactively.

I could add a unit test for just the bit of code that I changed, but it wouldn't really test much other than the keyword argument was passed to pyscopg2. I would rather handle the unit tests of all the scrapers separately. I will file a tech debt ticket for this.

What do you think?

lioramilbaum:
> > Can you please provide (at least) unit test?
> 
> Unfortunately, the scrapers don't have unit tests since they were only written to be used temporarily until Estuary-Updater was written shortly after. The scrapers have stayed operational to account for message bus outages and/or Estuary-Updater bugs, but unit tests were never added retroactively.
> 
> I could add a unit test for just the bit of code that I changed, but it wouldn't really test much other than the keyword argument was passed to pyscopg2. I would rather handle the unit tests of all the scrapers separately. I will file a tech debt ticket for this.
> 
> What do you think?

I get that there is a technical dept here.  Understood. On the other hand, nobody likes implementing tests. Such a ticket will stay in the backlog for ever until someone deletes it.

Can you implement a unit test only for the function you have touched? 
The unit tests coverage would increase by no time as soon as we embrace this practice.
WDYT?

mprahl:
I'm not going to get around to finishing this. Feel free to merge it as is or take over the PR.

PR #316, opened 2023-07-24
Including URL CA list - CWFHEALTH-1947

celeztyne:
Adding a new way to create the bundle certificate file to include new CA root. The old can't be excluded because is still valid (till 2024).  To implement this there is a new var CA_URLS, that is a list of root certificates separated by comma that afterwards will be added to a file


celeztyne:
@qixiang

qixiang:
We don't have capacity to fix the broken tests, will just merge this with CI failures.

